PR #31, opened 2022-06-04
DOCSP-22963 create quickstart page

davemungo:
NOTE: In the interest of time, sending out for internal and external review together. 

[STAGING](https://docs-mongodbcom-staging.corp.mongodb.com/cluster-sync/docsworker-xlarge/DOCSP-22963-create-quickstart-page/quickstart/)

[JIRA](https://jira.mongodb.org/browse/DOCSP-22963)


source/quickstart.txt
```diff
@@ -1,12 +1,235 @@
...
+~~~~~~~~~~~~~~~~~~~~~
+
+- The default port for the REST API is ``27182``.  Use the ``--port``
+  option with ``mongosync`` to :ref:`configure another port
+  <c2c-mongosync>`
+
+- ``mongosync`` can swap the source and destination clusters to enable
+  :ref:`reverse synchronization <c2c-api-reverse>`. To use ``reverse
+  synchronization``, you must pass these additional fields when you
+  start ``mongosync``:
```
alan-zheng-mdb, line 228: include link to /start endpoint

source/quickstart.txt
```diff
@@ -1,12 +1,235 @@
...
+Setup
+-----
+
+{+c2c-product-name+} synchronizes data between two clusters. To run
+``mongosync``, you will need to: 
+
+- Define a source and a destination cluster.
+- Define administrative users 
+- Download and install ``mongosync``.
+- Connect the clusters.
```
alan-zheng-mdb, line 35: suggestion: also label these as step 1, 2, 3, 4

source/quickstart.txt
```diff
@@ -1,12 +1,235 @@
...
+      :ref:`c2c-mongosync` is the tool that connects the source and
+      destination clusters. You can host ``mongosync`` on its own
+      hardware, ``mongosync`` does not have to run on the hardware that
+      hosts your MongodDB clusters.
+
+      To install ``mongosync``:
+      
+      #. Locate a host server for the ``mongosync`` executable that has
+         network connectivity to your source and destination clusters.
+      #. Go to the {+mdb-download-center+}.
```
alan-zheng-mdb, line 113: can this redirect to https://www.mongodb.com/try/download/mongosync. This will be live 8AM Tuesday Jun 7.

Please also double-check the ones in the installation page

snooty.toml
```diff
@@ -26,6 +26,7 @@ toc_landing_pages = ["/quickstart",
 version = "1.0.0"
 c2c-product-name = "Cluster-to-Cluster Sync"
 c2c-full-product-name = "MongoDB Cluster-to-Cluster Sync"
+mdb-download-center = "`MongoDB Download Center <https://www.mongodb.com/try/download/enterprise?tck=docs_server>`__"
```
alan-zheng-mdb, line 29: https://www.mongodb.com/try/download/mongosync

source/quickstart.txt
```diff
@@ -1,12 +1,235 @@
...
+      as your source cluster. If the source cluster is :ref:`sharded
+      <sharding-background>`, the destination cluster must be sharded
+      in the same way. If the source cluster is a :ref:`replica set
+      <replication>` with 3 nodes, the destination cluster must also be
+      a replica set with 3 nodes.
+
+      The source and destination clusters must be at least MongoDB 6.0.
+      If your clusters are self-managed, they must be MongoDB
+      Enterprise clusters. {+c2c-product-name+} is not supported on
+      MongoDB Community Edition.
```
alan-zheng-mdb, line 59: We do support MongoDB Community Edition, but our license only allows for customers migrating from community -> Atlas. See similar comment in https://github.com/mongodb/docs-cluster-to-cluster-sync/pull/32

source/quickstart.txt
```diff
@@ -1,12 +1,235 @@
...
+      hosted in :atlas:`MongodDB Atlas </getting-started?jmp=docs>`, 
+      use that cluster as the source cluster. If you don't have a
+      cluster to work with, you will need to create one.
+
+      You will need a destination cluster that has the same topology
+      as your source cluster. If the source cluster is :ref:`sharded
+      <sharding-background>`, the destination cluster must be sharded
+      in the same way. If the source cluster is a :ref:`replica set
+      <replication>` with 3 nodes, the destination cluster must also be
+      a replica set with 3 nodes.
```
alan-zheng-mdb, line 54: Suggestion flipping the statements so that the replica set goes first. 

The replica set statement is not true: 
If the source cluster is a replica set with 3 nodes, the destination cluster must also be a replica set but can have the same or a different number of nodes.

alan-zheng-mdb (changes requested):
Thanks Dave. Some corrections to a few statements

source/quickstart.txt
```diff
@@ -1,12 +1,235 @@
...
+      - :authrole:`clusterMonitor` role
+      - :authrole:`backup` role
+
+      In addition, the source user must be able to:
+
+      - Run the :dbcommand:`getParameter` command
+
+      If the source cluster is hosted in Atlas, the user must have the
+      :ref:`Atlas admin <atlas-user-privileges>` role. The user must
+      also be able to :ref:`read the oplog <ref-atlas-oplog>`.
```
ldennis, line 87: `mongosync` uses change streams, not oplog. Should we mention something like [this](https://www.mongodb.com/docs/v6.0/changeStreams/#access-control) here? I think `mongosync` would read the change stream for the entire cluster.

source/quickstart.txt
```diff
@@ -1,12 +1,235 @@
...
+      permissions in both clusters. 
+
+      Source Cluster Authentication Requirements
+      ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+      The source user must have the following roles:
+
+      - :authrole:`readAnyDatabase` role
+      - :authrole:`clusterMonitor` role
+      - :authrole:`backup` role
```
ldennis, line 79: I will need to double check with the engineering team tomorrow.

source/quickstart.txt
```diff
@@ -1,12 +1,235 @@
...
+  :ref:`start <c2c-api-start>` command. 
+- Designate the source and destination clusters with the :ref:`start
+  <c2c-api-start>` command. "cluster0" and "cluster1" are just labels,
+  either cluster can be ``cluster0`` or ``cluster1``.
+
+Synchronize Data Between Clusters
+---------------------------------
+
+The command interface for ``mongosync`` is an HTTP server that
+publishes a REST API. To control ``mongosync``, use the API endpoints.
```
tfogo, line 180: Nit: technically it's not a REST API. We should just refer to it as an HTTP API.

source/quickstart.txt
```diff
@@ -1,12 +1,235 @@
...
+      For operating system specific installation instructions, see
+      :ref:`c2c-install`.
+
+   .. step:: Connect the Clusters
+
+      A connection string contains the network and authentication
+      details that ``mongosync`` needs to connect to the source and
+      destination clusters.
+
+      Determine the IP address and port for your source and destination
```
tfogo, line 127: Nit: I'd say "hostname or IP address"

tfogo (commented):
This is looking good. I know we're tight on time, but if we can, I think we should add two more sections. I think ideally the quick start should also explain how to poll the progress API and how to commit. Without explaining how to do a cutover, we're not fully explaining how to use the tool. We could probably just copy the example from the commit endpoint reference.

source/quickstart.txt
```diff
@@ -1,12 +1,235 @@
...
+------------------------
+
+``monogsync`` must create an initial connection to the source and
+destination clusters before it can start to sync data. To create the
+initial connection, issue the following command on a single line (the
+command is reformated here for clarity):
+
+.. code-block:: shell
+
+   ./bin/mongosync --cluster0 mongodb://clusterAdmin:superSecret@clusterOne01.fancyCorp.com:20020,
```
tfogo, line 158: Making this multi-line means we can't copy/paste this into a terminal. Either make it one line, or include `\` line breaks where appropriate.

source/quickstart.txt
```diff
@@ -1,12 +1,235 @@
...
+---------------------------------
+
+The command interface for ``mongosync`` is an HTTP server that
+publishes a REST API. To control ``mongosync``, use the API endpoints.
+
+The :ref:`start <c2c-api-start>` command initiates data synchronization.
+To start synchronizing, use ``curl`` or a similar program to issue the
+:ref:`start <c2c-api-start>` command:
+
+.. literalinclude:: /includes/api/requests/start.sh
```
tfogo, line 186: Same issue here with copying into a terminal. Either should be 1 line, or add a `\` at the end of the first line.

source/quickstart.txt
```diff
@@ -1,12 +1,235 @@
...
+   * - :ref:`c2c-api-resume`
+     - Resumes a paused synchronization session based on data stored on
+       the destination cluster.
+   * - :ref:`c2c-api-reverse`
+     - Reverses the direction of a committed sync operation.
+   * - :ref:`c2c-api-start`
+     - Starts the synchronization between a source and destination
+       cluster.
+
+Synchronization Notes
```
tfogo, line 218: I think it might be good to mention that users may need to increase the file descriptor ulimits when running mongosync (particularly for macos which has low defaults). We could link to https://www.mongodb.com/docs/v5.3/reference/ulimit/

I don't know if it's best here or in some other section, but I think people testing on macos are likely to run into this problem so it's worth mentioning.

source/quickstart.txt
```diff
@@ -1,12 +1,235 @@
...
+     - Commits the synchronization operation to the destination
+       cluster.
+   * - :ref:`c2c-api-pause`
+     - Pauses the current synchronization operation.
+   * - :ref:`c2c-api-progress`
+     - Returns the status of the synchronization process.
+   * - :ref:`c2c-api-resume`
+     - Resumes a paused synchronization session based on data stored on
+       the destination cluster.
+   * - :ref:`c2c-api-reverse`
```
ksuarz, line 212: Just wanted to mention that the rendering on staging for this option looked wrong:
Screen Shot 2022-06-06 at 10 52 45 AM

davemungo: Thanks - I think the staging link is out of date. I'll update it

source/quickstart.txt
davemungo: Done

source/quickstart.txt
```diff
@@ -1,12 +1,252 @@
...
+
+      You will need a destination cluster that has the same topology
+      as your source cluster:
+      
+      - If the source cluster is a :ref:`replica  set <replication>`,
+        the destination cluster must also be a replica set. The number
+        of nodes in the destination replica set does not have to equal
+        the number of nodes in the source replica set.
+      - If the source cluster is :ref:`sharded  <sharding-background>`,
+        the destination cluster must have the same number of shards.
```
alan-zheng-mdb, line 57: same number of shards and the same shard key. 

https://www.mongodb.com/docs/manual/core/sharding-shard-key/

source/quickstart.txt
alan-zheng-mdb: Understand that this is alphabetized, but /commit as the first API threw me off. can we match the ordering under the API dropdown? 

under the reference -> API page we followed the order:
/start
/progress
/pause
/resume
/commit
/reverse

source/quickstart.txt
davemungo: Done

source/quickstart.txt
davemungo: Done

davemungo (commented):
Thanks @alan-zheng-mdb I've update the copy. 

Merging next

PR #41, opened 2022-06-17
DOCSP-22863 Configuration 

kennethdyer:
[DOCSP-22863](https://jira.mongodb.org/browse/DOCSP-22863)

STAGING:
* [Configuration File Settings](https://docs-mongodbcom-staging.corp.mongodb.com/cluster-sync/docsworker-xlarge/DOCSP-22863-configuration/reference/configuration/)
* [Configuration File Settings to Command-line Option Mapping](https://docs-mongodbcom-staging.corp.mongodb.com/cluster-sync/docsworker-xlarge/DOCSP-22863-configuration/reference/configuration/file-setting-cli-option-mapping/)

Also, refactors [mongosync CLI options](https://docs-mongodbcom-staging.corp.mongodb.com/cluster-sync/docsworker-xlarge/DOCSP-22863-configuration/reference/mongosync/) to use includes.


kennethdyer:
@alan-zheng-mdb We could feature a better example.  I'd suggest we make a point of it being basic example rather than a full example.  Like, `cluster%`, `logPath`, and maybe `verbosity: WARN`.  That way if someone copypastas from the Docs they don't pick up too many unnecessary settings.

Might also have a code block under individual settings showing an example of it in use.

source/reference/configuration/file-setting-cli-option-mapping.txt
```diff
@@ -0,0 +1,29 @@
+.. _c2c-config-setting-option-map:
+
+Configuration File Settings and Command-Line Option Mapping
```
ianf-mongodb, line 3: Since the primary use of this page is the Command-Line reference, can you update the title to be "Command-Line Options".

source/reference/configuration.txt
```diff
@@ -0,0 +1,110 @@
...
+For example:
+
+.. code-block:: yaml
+
+   cluster0: "mongodb://192.0.2.10:27017"
+   cluster1: "mongodb://192.0.2.20:27017"
+   logPath: "/var/log/mongodb/mongosync"
+   verbosity: "WARN"
+
+To use a configuration file, start ``mongosync`` with the :option:`--config`
```
ianf-mongodb, line 35: Just curious if this connection needs to specified toward a `mongos` instance. If so might need to add mention of this.

ianf-mongodb (commented):
@kennethdyer This looks really good, I left a couple of comments for consideration. Thanks!

source/reference/configuration.txt
kennethdyer: Reference was meant as generic to MongoDB, switched it to `/var/log/mongosync/` to avoid confusion.

source/reference/configuration.txt
```diff
@@ -0,0 +1,110 @@
+.. _c2c-config:
+
+Configuration File Settings
+###########################
+
+.. include:: /includes/preview-warning
```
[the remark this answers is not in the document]
kennethdyer: The use of includes copies other pages in C2C.  Per Dave, includes were chosen over banners due to formatting issues encountered during the initial setup.

source/reference/configuration/file-setting-cli-option-mapping.txt
```diff
@@ -0,0 +1,29 @@
+.. _c2c-config-setting-option-map:
```
alan-zheng-mdb, line 1: do we really need this sub-page? It feels repetitive as it is already mentioned in the parent config file page

source/reference/configuration.txt
```diff
@@ -0,0 +1,110 @@
...
+
+
+Configuration File
+==================
+
+Most :program:`mongosync` command-line options can be written to file.
+The configuration file specifies values for each setting using YaML format.
+For example:
+
+.. code-block:: yaml
```
alan-zheng-mdb, line 28: now that we have expanded the config file example with it's own page, do we still need this code-block example here? In this section, should we also just link to that page for more information?

kennethdyer: @alan-zheng-mdb, Could you clarify on this point?  I don't see a page with an expanded config file example in the build.

source/reference/configuration.txt
```diff
@@ -0,0 +1,105 @@
...
+You can configure :program:`mongosync` instances at startup using a configuration
+file.  The configuration file contains settings that are the equivalent of
+``mongosync`` command-line options.
+
+
+Configuration File
+==================
+
+Most :program:`mongosync` command-line options can be written to file.
+The configuration file specifies values for each setting using YaML format.
```
tfogo, line 25: YAML is usually all-caps.

tfogo (changes requested):
Looking good. It did seem a bit weird to me that the `--config` option links to the small description on the same page instead of the new page: https://docs-mongodbcom-staging.corp.mongodb.com/cluster-sync/docsworker-xlarge/DOCSP-22863-configuration/reference/mongosync/#std-option-mongosync.--config

But wither way is probably fine.

source/reference/configuration.txt
kennethdyer: Fixed

source/reference/configuration.txt
kennethdyer: Fixed

kennethdyer:
@tfogo Content updated per your review.

The `--config` option link works as intended.  We use a different syntax when linking to the overall Configuration page.

tfogo (approved):
LGTM!

PR #47, opened 2022-06-29
DOCS-15126-conflict-resolution-during-sharding

davemungo:
[STAGING](https://docs-mongodbcom-staging.corp.mongodb.com/cluster-sync/docsworker-xlarge/DOCS-15126-conflict-resolution-for-sharded-clusters-v0.9.0/reference/mongosync/#sharded-clusters)

[JIRA](https://jira.mongodb.org/browse/DOCS-15126)

### DIFF
--- a/source/reference/mongosync.txt
+++ b/source/reference/mongosync.txt
@@ -170,10 +170,9 @@ the destination cluster, however a :ref:`chunk migration
 <sharding-chunk-migration>` or similar source update could move
 documents to a new source shard during replication.
 
-To preserve event order on the destination cluster,
-{+c2c-product-name+} automatically adjusts the :ref:`change events
-<change-events>` if documents move between source shards during
-replication.
+Even if documents move between source shards during replication,
+{+c2c-product-name+} maintains the :term:`eventual consistency`
+guarantee on the destination cluster.
 
 Views
 ~~~~~
--- a/source/reference/mongosync.txt
+++ b/source/reference/mongosync.txt
@@ -170,10 +170,10 @@ the destination cluster, however a :ref:`chunk migration
 <sharding-chunk-migration>` or similar source update could move
 documents to a new source shard during replication.
 
-If a document moves from one source shard to another during
-replication, {+c2c-product-name+} automatically adjusts the
-:ref:`change events <change-events>` to preserve event order on the
-destination cluster.
+To preserve event order on the destination cluster,
+{+c2c-product-name+} automatically adjusts the :ref:`change events
+<change-events>` if documents move between source shards during
+replication.
 
 Views
 ~~~~~
--- a/source/reference/mongosync.txt
+++ b/source/reference/mongosync.txt
@@ -161,6 +161,20 @@ source database is dropped at a later time. The ``<db>.system.profile``
 collection will not change the accuracy of user data on the
 destination.
 
+Sharded Clusters
+~~~~~~~~~~~~~~~~
+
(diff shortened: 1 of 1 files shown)


source/reference/mongosync.txt
```diff
@@ -161,6 +161,20 @@ source database is dropped at a later time. The ``<db>.system.profile``
...
+Sharded Clusters
+~~~~~~~~~~~~~~~~
+
+{+c2c-product-name+} supports replication between sharded clusters.
+Individual shards are replicated in parallel from the source cluster to
+the destination cluster, however a :ref:`chunk migration
+<sharding-chunk-migration>` or similar source update could move
+documents to a new source shard during replication.
+
+If a document moves from one source shard to another during
```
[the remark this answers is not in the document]
davemungo: The PM notes refer to chunk migration as an example of moving a document between shards, I'm not sure if there are other mechanisms too. 

I can make documents plural though, which may address your suggestion.

davemungo (commented):
Thanks for the feedback @ianf-mongodb 

davemungo:
@alan-zheng-mdb would you take a look at this one when you have a moment? Thanks!

alan-zheng-mdb:
@ldennis can you take this one?

source/reference/mongosync.txt
```diff
@@ -161,6 +161,20 @@ source database is dropped at a later time. The ``<db>.system.profile``
...
+~~~~~~~~~~~~~~~~
+
+{+c2c-product-name+} supports replication between sharded clusters.
+Individual shards are replicated in parallel from the source cluster to
+the destination cluster, however a :ref:`chunk migration
+<sharding-chunk-migration>` or similar source update could move
+documents to a new source shard during replication.
+
+To preserve event order on the destination cluster,
+{+c2c-product-name+} automatically adjusts the :ref:`change events
```
ldennis, line 174: Not quite sure what "adjusts" means here. But I assume this is fine. This is implementation details of how we handle out of order events while maintaining eventual consistency. I guess the end users don't have to know. So this is probably fine then.

Just want to provide more context here: https://jira.mongodb.org/browse/REP-33 is the project to implement the core replication machinery for sharded clusters. I don't think there is much we need to document here other than a statement that we support sharded clusters. So I think the level of details we have here is probably fine.

source/reference/mongosync.txt
```diff
@@ -161,6 +161,20 @@ source database is dropped at a later time. The ``<db>.system.profile``
...
+Sharded Clusters
+~~~~~~~~~~~~~~~~
+
+{+c2c-product-name+} supports replication between sharded clusters.
+Individual shards are replicated in parallel from the source cluster to
+the destination cluster, however a :ref:`chunk migration
+<sharding-chunk-migration>` or similar source update could move
+documents to a new source shard during replication.
+
+To preserve event order on the destination cluster,
```
ldennis, line 173: We don't "preserve" event order. In fact, we are eventually consistent meaning that the write history/order may not be the same and we only guarantee a consistent snapshot after a healthy cutover. So I would avoid saying "preserve". Instead, I think we just have a mechanism to handle it correctly while maintaining the eventually consistent guarantee.

source/reference/mongosync.txt
davemungo: Acknowledged

ldennis (approved):

davemungo (commented):
Thanks @ldennis  this is ready for another pass

PR #48, opened 2022-07-01
DOCS-14974 basic support for sharded

davemungo:
[STAGING](https://docs-mongodbcom-staging.corp.mongodb.com/cluster-sync/docsworker-xlarge/DOCS-14974-basic-support-for-sharded-v0.9.0/multiple-mongosyncs/)

[JIRA](https://jira.mongodb.org/browse/DOCS-14974)


source/reference/multiple-mongosyncs.txt
```diff
@@ -0,0 +1,232 @@
...
+
+Step 3. Connect the ``mongosync`` instances
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+These instructions use a generic connection string. To modify the
+connection string for your cluster architecture, refer to the
+architecture specific :ref:`connection details <c2c-connecting>`.
+
+A single host server can run multiple ``mongosync`` instances. To
+improve performance, run ``mongosync`` on multiple host servers.
```
jeff-allen-mongo, line 86: Consider putting this paragraph in a note or tip admonition.

source/reference/multiple-mongosyncs.txt
```diff
@@ -0,0 +1,232 @@
...
+        '{ "source": "cluster0", "destination": "cluster1", \
+           "reversible": false, "enableUserWriteBlocking": false }'
+
+   curl monogsync02Host:27602/api/v1/start -XPOST --data \
+        '{ "source": "cluster0", "destination": "cluster1", \
+           "reversible": false, "enableUserWriteBlocking": false }'
+
+The ``start`` command options, in this example the ``--data`` values in
+the ``curl`` command, must be the same for all of the ``mongosync``
+instances.
```
jeff-allen-mongo, line 139: This sentence took me a few reads to understand. I would omit "in this example the ``--data`` values in the ``curl`` command" to make this a bit easier to parse.

source/reference/multiple-mongosyncs.txt
```diff
@@ -0,0 +1,232 @@
...
+
+.. _c2c-sharded-progress:
+
+Check Progress on a Single ``mongosync`` Instance
```
jeff-allen-mongo, line 143: I'm having a hard time understanding what this section and the ones following it have to do with Sharded Clusters specifically.

jeff-allen-mongo (changes requested):
@davemungo I left some comments. I'm kind of missing what a lot of these sections on this page have to do with sharded clusters, as they seem to be discussing single mongosync instances outside of the context of a sharded cluster.

davemungo (commented):
Thanks for your feedback @jeff-allen-mongo 

source/multiple-mongosyncs.txt
```diff
@@ -0,0 +1,255 @@
...
+             },
+             {
+               _id: 'shard02',
+               host: 'shard02/localhost:27504,localhost:27505,localhost:27506',
+               state: 1,
+               topologyTime: Timestamp({ t: 1656612240, i: 4 })
+             }
+           ]
+
+   .. step:: 3. Connect the ``mongosync`` instances
```
jeff-allen-mongo, line 92: This step doesn't appear to show up in staging for me?

jeff-allen-mongo (commented):
@davemungo thanks for making these edits. The intro now helps me understand the goal of the page, and I think using the `.. procedure::` syntax makes the flow easier to follow.

I've left just a couple more comments for your consideration.

Jeff

source/multiple-mongosyncs.txt
davemungo: The staging link was stale. It's updated now.

davemungo (commented):
Thanks for the feedback @jeff-allen-mongo 

jeff-allen-mongo (approved):
LGTM

source/multiple-mongosyncs.txt
```diff
@@ -0,0 +1,255 @@
...
+           "reversible": false, "enableUserWriteBlocking": false }'
+
+   curl monogsync02Host:27602/api/v1/start -XPOST --data \
+        '{ "source": "cluster0", "destination": "cluster1", \
+           "reversible": false, "enableUserWriteBlocking": false }'
+
+The ``start`` command options must be the same for all of the ``mongosync``
+instances.
+
+.. _c2c-sharded-progress:
```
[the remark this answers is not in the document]
davemungo: I had a look at the API docs, there isn't much duplication of their content. 

The 'this works on one shard, repeat on the others' part does get repeated in various forms. We could move it to the beginning of the section, but since these aren't steps we don't know how a reader arrives at a given procedure. They could easily miss a note that's farther up the page.

We could remove the example like this:


Check Progress
--------------

The ``mongosync`` instances do not aggregate progress information
across shards. To review synchronization progress for a particular
shard, use ``curl`` or another HTTP client to issue the
:ref:`progress <c2c-api-progress>` command to the ``mongosync``
instance synchronizing that shard. To check progress on other shards repeat
the API call to each ``mongosync`` instance.

But then we force the reader to jump to another page. Leaving the example in gives the user the basic code they need.

davemungo (commented):
I'll add the question about whether or not we need to say much about the single mongosync use case to the other [PR](https://github.com/mongodb/docs-cluster-to-cluster-sync/pull/57)

PR #50, opened 2022-07-05
DOCSP-23597 Remove limitations

davemungo:
[STAGING](https://docs-mongodbcom-staging.corp.mongodb.com/cluster-sync/docsworker-xlarge/DOCSP-23597-remove-limitations-v1.0/reference/limitations/)

[JIRA](https://jira.mongodb.org/browse/DOCSP-23597)

### DIFF
--- a/source/reference/limitations.txt
+++ b/source/reference/limitations.txt
@@ -29,7 +29,6 @@ General Limitations
 - The source and destination clusters must have the same Feature
   Compatibility Version.
 - The destination cluster must be empty.
-- DDL operations are currently not supported during syncing. 
 - ``mongosync`` does not validate that the clusters or the environment
   are properly configured. 
 - The primary shard cannot be :dbcommand:`reassigned <movePrimary>`


source/reference/limitations.txt
```diff
@@ -65,9 +64,7 @@ discuss your requirements.
 Unsupported Collection Types
```
jeff-allen-mongo, line 64: Now that there is only one unsupported collection type, consider making this heading more specific:

```suggestion
Time Series Collections
```

jeff-allen-mongo (approved):
LGTM with a suggestion

source/reference/limitations.txt
jeff-allen-mongo: Actually I see in PR #51 that we're adding another bullet to this section, so this can stay as-is.

davemungo:
@ldennis  would you mind giving this one review since Alan is out?

ldennis (approved):
LGTM

PR #52, opened 2022-07-06
DOCSP-23350 write-blocking note

davemungo:
[STAGING](https://docs-mongodbcom-staging.corp.mongodb.com/cluster-sync/docsworker-xlarge/DOCSP-23350-write-blocking-v0.9.0/reference/mongosync/#write-blocking)

[JIRA](https://jira.mongodb.org/browse/DOCSP-23350)

### DIFF
--- a/source/reference/mongosync.txt
+++ b/source/reference/mongosync.txt
@@ -94,17 +94,6 @@ Global Options
    Sets the port used by the HTTP server for the {+c2c-product-name+}
    HTTP API.
 
-.. option:: --reversible
-
-   *Default*: false
-
-   To :ref:`reverse the direction <c2c-api-reverse>` of synchronization
-   between the source and destination clusters, ``mongosync`` must be
-   started with ``--reversible true``. 
-
-   If ``--reversible`` is set to ``true``, then
-   ``enableUserWriteBlocking`` must also be set to ``true``.
-
 .. option:: --verbosity <level>
 
    *Default*: ``INFO``
@@ -207,11 +196,13 @@ To enable write-blocking, start ``mongosync`` with
 
 :ref:`Reverse synchronization <c2c-api-reverse>` requires
 write-blocking. To enable reverse synchronization between the source
-and destination clusters, start ``mongosync`` with ``reversible`` and
-``enableUserWriteBlocking`` set to ``true``.
+and destination clusters, set ``reversible`` and
+``enableUserWriteBlocking`` to ``true`` when you use the
+:ref:`start API <c2c-api-start>` to initiate ``mongosync``. The options
+must be set at startup, they cannot be updated later.
 
 To set ``enableUserWriteBlocking``, the ``mongosync`` user must have a
-role that includes the ``setUserWriteBlock`` and
+role that includes the ``setUserWriteBlockMode`` and
 ``bypassWriteBlockingMode`` ActionTypes.
 
 Examples
--- a/source/reference/mongosync.txt
+++ b/source/reference/mongosync.txt
@@ -200,7 +200,7 @@ By default, write-blocking is not enabled during synchronization. When
 write-blocking is enabled, ``mongosync`` blocks writes:
 
 - On the destination cluster during replication
-- On the source cluster blocks while committing
+- On the source cluster while committing
 
 To enable write-blocking, start ``mongosync`` with
 ``enableUserWriteBlocking`` set to ``true``.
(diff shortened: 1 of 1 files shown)


source/reference/mongosync.txt
```diff
@@ -182,6 +193,27 @@ show an empty ``system.views`` collection in that database. The empty
...
 data on the destination.
 
+Write Blocking
+~~~~~~~~~~~~~~
+
+By default, write-blocking is not enabled during synchronization. When
+write-blocking is enabled, ``mongosync`` blocks writes:
+
+- On the destination cluster during replication
+- On the source cluster blocks while committing
```
jmd-mongo, line 203: [extra word]

I think "blocks" on line 203 can be removed.

davemungo (commented):
Thanks for the feedback @jmd-mongo this is ready for another pass.

davemungo:
@ldennis this is one of the carry over notes from MDB World. Would you have  look?

source/reference/mongosync.txt
```diff
@@ -91,30 +94,38 @@ Global Options
    Sets the port used by the HTTP server for the {+c2c-product-name+}
    HTTP API.
 
-.. option:: --id <ID>
+.. option:: --reversible
```
ldennis, line 97: This is not a command line option we support. Instead, `reversible` is an option for the `/start` API.

source/reference/mongosync.txt
```diff
@@ -182,6 +193,27 @@ show an empty ``system.views`` collection in that database. The empty
...
+~~~~~~~~~~~~~~
+
+By default, write-blocking is not enabled during synchronization. When
+write-blocking is enabled, ``mongosync`` blocks writes:
+
+- On the destination cluster during replication
+- On the source cluster while committing
+
+To enable write-blocking, start ``mongosync`` with
+``enableUserWriteBlocking`` set to ``true``.
```
ldennis, line 206: I think this should link to the `/start` API as `enableUserWriteBlocking` an option there.

source/reference/mongosync.txt
```diff
@@ -182,6 +193,27 @@ show an empty ``system.views`` collection in that database. The empty
...
+
+:ref:`Reverse synchronization <c2c-api-reverse>` requires
+write-blocking. To enable reverse synchronization between the source
+and destination clusters, start ``mongosync`` with ``reversible`` and
+``enableUserWriteBlocking`` set to ``true``.
```
ldennis, line 211: I would also mention that once `reversible` / `enableUserWriteBlocking` is set, they are not allowed to change. And users are only able to config them the value of these two options at the time of `/start` (e.g. not at the time of`/reverse`).

source/reference/mongosync.txt
```diff
@@ -182,6 +193,27 @@ show an empty ``system.views`` collection in that database. The empty
...
+
+To set ``enableUserWriteBlocking``, the ``mongosync`` user must have a
+role that includes the ``setUserWriteBlock`` and
```
ldennis, line 214:
```suggestion
role that includes the ``setUserWriteBlockMode`` and
```

ldennis (changes requested):

davemungo (commented):
Thanks for your feedback @ldennis 

source/reference/mongosync.txt
ldennis: > start mongosync with enableUserWriteBlocking set to true.

I still think this is misleading to the end users. It is not clear if this means start of the mongosync process or start of the replication. Since this is a /start API, we meant start of the replication.

source/reference/mongosync.txt
```diff
@@ -182,6 +182,29 @@ show an empty ``system.views`` collection in that database. The empty
...
+
+To enable write-blocking, start ``mongosync`` with
+``enableUserWriteBlocking`` set to ``true``.
+
+:ref:`Reverse synchronization <c2c-api-reverse>` requires
+write-blocking. To enable reverse synchronization between the source
+and destination clusters, set ``reversible`` and
+``enableUserWriteBlocking`` to ``true`` when you use the
+:ref:`start API <c2c-api-start>` to initiate ``mongosync``. The options
+must be set at startup, they cannot be updated later.
```
ldennis, line 202: ditto, I still think `startup` is a bit unclear.

davemungo: Done

davemungo (commented):
@ldennis thanks for your feedback

ldennis (approved):
LGTM, thank you!

PR #54, opened 2022-07-06
DOCSP-23407 rename and drop not supported

davemungo:
[STAGING](https://docs-mongodbcom-staging.corp.mongodb.com/cluster-sync/docsworker-xlarge/DOCSP-23407-rename-drop-not-supported-v1.0.0/reference/limitations/)

[JIRA](https://jira.mongodb.org/browse/DOCSP-23407)

### DIFF
--- a/source/reference/limitations.txt
+++ b/source/reference/limitations.txt
@@ -53,6 +53,9 @@ General Limitations
 - `Queryable Execution
   <https://www.mongodb.com/docs/v6.0/core/queryable-encryption/>`__ is
   not supported.
+- If the :dbcommand:`renameCollection` or the :dbcommand:`drop` command
+  modifies a :ref:`capped collection <manual-capped-collection>`,
+  synchronization is not supported.
 
 MongoDB Community Edition
 -------------------------


jeff-allen-mongo (commented):
@davemungo thanks for this addition. I've left a question because I'm not totally clear on the behavior here.

source/reference/limitations.txt
```diff
@@ -53,6 +53,9 @@ General Limitations
 - `Queryable Execution
   <https://www.mongodb.com/docs/v6.0/core/queryable-encryption/>`__ is
   not supported.
+- If the :dbcommand:`renameCollection` or the :dbcommand:`drop` command
+  modifies a :ref:`capped collection <manual-capped-collection>`,
+  synchronization is not supported.
```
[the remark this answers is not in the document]
davemungo: It seems there could be a couple of outcomes. The jira ticket says there may be some empty collections created (which is regarded as inconsistency) and/or mongosync could crash.

source/reference/limitations.txt
```diff
@@ -53,6 +53,9 @@ General Limitations
 - `Queryable Execution
   <https://www.mongodb.com/docs/v6.0/core/queryable-encryption/>`__ is
   not supported.
+- If the :dbcommand:`renameCollection` or the :dbcommand:`drop` command
+  modifies a :ref:`capped collection <manual-capped-collection>`,
```
jeff-allen-mongo, line 57: Using "modify" to refer to "drop" seems a little misleading. What about:

"If you run the :dbcommand:`renameCollection` or the :dbcommand:`drop` command on a :ref:`capped collection <manual-capped-collection>`..."

source/reference/limitations.txt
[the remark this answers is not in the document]
jeff-allen-mongo: In that case, consider adding the following clarification:

"... synchronization is not supported for the cluster containing the capped collection."

This may help users understand that it's not just the one collection that's affected, but potentially the entire cluster.

jeff-allen-mongo (approved):
LGTM % suggestions. Thanks @davemungo

davemungo (commented):
Thanks @jeff-allen-mongo  sending to external review next

davemungo:
@WenbinZhu when you have a moment, would you review this docs update for C2C?

source/reference/limitations.txt
```diff
@@ -53,6 +53,10 @@ General Limitations
 - `Queryable Execution
   <https://www.mongodb.com/docs/v6.0/core/queryable-encryption/>`__ is
   not supported.
+- If you run the :dbcommand:`renameCollection` command or the
+  :dbcommand:`drop` command on a :ref:`capped collection
+  <manual-capped-collection>`, synchronization is not supported on the
+  cluster.
```
WenbinZhu, line 59: I think to be more precise, we can say it would be undefined behavior to do so?

WenbinZhu (approved):
LGTM! Sorry for the delay, I just have one nit comment

source/reference/limitations.txt
davemungo: Done

PR #57, opened 2022-07-07
DOCS-14974 basic support for sharded v0.9.0

davemungo:
[STAGING](https://docs-mongodbcom-staging.corp.mongodb.com/cluster-sync/docsworker-xlarge/DOCS-14974-basic-support-for-sharded-v0.9.0/multiple-mongosyncs/)

[JIRA](https://jira.mongodb.org/browse/DOCS-14974)

This was originally reviews as [PR 48](https://github.com/mongodb/docs-cluster-to-cluster-sync/pull/48)


source/multiple-mongosyncs.txt
```diff
@@ -0,0 +1,260 @@
...
+
+There are two ways to synchronize :ref:`sharded clusters
+<sharded-cluster>`. You can use either one ``monogosync`` or several
+``monogosync`` instances. For best performance, use multiple
+``monogosync`` instances, one ``monogosync`` for each shard in the
+cluster.
+
+.. _c2c-sharded-config-single:
+
+Configure a Single ``mongosync`` Instance
```
davemungo, line 25: @huayu-ouyang  I'm not sure what the use cases are for running a single mongosync. Is it better to demote the single monogsync method to a note or should it be included along with the multiple monogsync method?

source/multiple-mongosyncs.txt
```diff
@@ -0,0 +1,260 @@
...
+instance that is running on ``monogsync01Host`` and using ``port
+27601``. Make additional calls to ``commit`` on any other
+``mongosync`` instances that may be running.
+
+.. _c2c-sharded-reverse:
+
+Reverse the Synchronization Direction
+-------------------------------------
+
+To reverse synchronization so that the original destination cluster
```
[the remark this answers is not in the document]
davemungo: Yes, it's in a couple of places but I can add a note here too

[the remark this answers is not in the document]
davemungo: Done.

davemungo (commented):
Thanks for the feedback @huayu-ouyang  

@alan-zheng-mdb is OOO. Do we need to wait for his feedback on your question?

source/multiple-mongosyncs.txt
davemungo: Thanks @ldennis i've modified the text to read: 

For best performance with large or heavily loaded clusters, use multiple ``monogosync`` instances,

And I'll leave the single monogsync case as is.

davemungo (commented):
Thanks for the feedback @huayu-ouyang  this is ready for another pass

source/multiple-mongosyncs.txt
```diff
@@ -0,0 +1,266 @@
...
+is running on ``monogsync01Host`` and using ``port 27601``.  To 
+resume synchronization on other shards, update the host and port number
+then repeat the API call to each ``mongosync`` instance.
+
+.. _c2c-sharded-commit:
+
+Commit Synchronization From  Multiple ``mongosync`` Instances
+-------------------------------------------------------------
+
+When synchronization is complete, issue the :ref:`commit
```
huayu-ouyang, line 223: just a question about this phrasing, not sure what the language is elsewhere in the docs but my understanding is that issuing the commit command is what completes the synchronization, and before issuing the commit command the synchronization is still ongoing, so what does "when synchronization is complete" mean in these docs?

huayu-ouyang (changes requested):

source/multiple-mongosyncs.txt
davemungo: @huayu-ouyang  I think this language came from the spec docs. 

@ldennis What does it mean for synchronization to be complete (pre-commit)?  Maybe another way to ask is, how can a user determine that it is time to send the /commit call to mongosync?

source/multiple-mongosyncs.txt
```diff
@@ -0,0 +1,266 @@
...
+This command only commits synchronization from the ``mongosync``
+instance that is running on ``monogsync01Host`` and using ``port
+27601``. Make additional calls to ``commit`` on any other
+``mongosync`` instances that may be running.
+
+.. _c2c-sharded-reverse:
+
+Reverse the Synchronization Direction
+-------------------------------------
+
```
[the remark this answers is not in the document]
davemungo: Done

davemungo (commented):
@ldennis, @huayu-ouyang 

There's an open question about when to commit. (line 223)

source/multiple-mongosyncs.txt
ldennis: Sorry, I think I missed that. I think it is totally up to the users when to issue /commit. They technically can commit whenever canCommit: true from the progress API. But normally, we would recommend issuing the commit API only when the lag time is small enough for their use case.

huayu-ouyang: Maybe instead of saying "when synchronization is complete" we could say something like "When you want to complete synchronization, check that canCommit: true, and then issue the commit command to each mongosync instance"? and we could add a note that we recommend only issuing the commit command when the lag time is small enough for their use case?

davemungo (commented):
@ldennis, @huayu-ouyang thanks for your feedback

huayu-ouyang (approved):
LGTM! sorry for the delayed review!

source/multiple-mongosyncs.txt
```diff
@@ -0,0 +1,281 @@
...
+-------------------------------------------------------------
+
+When you want to complete synchronization, issue the :ref:`progress
+<c2c-api-progress>` command. If the ``progress`` command returns,
+``{ canCommit: true }``, issue the :ref:`commit <c2c-api-commit>`
+command. Repeat the process on all of the ``mongosync`` instances.
+
+The ``commit`` operation is blocking. The ``commit`` command will not
+return until ``commit`` has been called on every ``mongosync``
+instance. To avoid unsynchronized data, you should only run the
```
ldennis, line 230: With a larger lag, it just mean the commit process would take longer. And the implication for customers is that they will have to block writes on the source for longer while mongosync is trying to drain the lag.

So I wouldn't say "To avoid unsynchronized data", this makes it sound like we could have data consistency problem if we commit with lag.

source/multiple-mongosyncs.txt
```diff
@@ -0,0 +1,281 @@
...
+then repeat the API call to each ``mongosync`` instance.
+
+.. _c2c-sharded-commit:
+
+Commit Synchronization From  Multiple ``mongosync`` Instances
+-------------------------------------------------------------
+
+When you want to complete synchronization, issue the :ref:`progress
+<c2c-api-progress>` command. If the ``progress`` command returns,
+``{ canCommit: true }``, issue the :ref:`commit <c2c-api-commit>`
```
ldennis, line 225: I think it would be better to pull the lag time consideration forward. {canCommit: true} means nothing but the initial cloning phase is done and that mongosync is tailing the source. I don't necessarily want users to have the impression that {canCommit: true} == ready to issue /commit.

ldennis (changes requested):

PR #60, opened 2022-07-08
DOCSP-23351 unique index req v0.9.0

davemungo:
[STAGING](https://docs-mongodbcom-staging.corp.mongodb.com/cluster-sync/docsworker-xlarge/DOCSP-23351-unique-index-req-v0.9.0/reference/api/reverse/#requirements)

[JIRA](https://jira.mongodb.org/browse/DOCSP-23351)

### DIFF
--- a/source/reference/api/reverse.txt
+++ b/source/reference/api/reverse.txt
@@ -40,8 +40,12 @@ To use the ``reverse`` endpoint:
 - You must have started your original sync operation with the
   ``reversible`` and ``enableUserWriteBlocking`` options set to
   ``true``.
-
 - ``mongosync`` must be in the ``COMMITTED`` state.
+- Any indexes in the cluster must be :ref:`unique indexes
+  <index-type-unique>`.You can =convert a non-unique index to a unique
+  index, with the :dbcommand:`collMod` command. See `this example
+  <https://www.mongodb.com/docs/v6.0/reference/command/collMod/#convert-an-existing-index-to-a-unique-index>`__
+  for more details.
 
 Request
 -------
--- a/source/reference/api/reverse.txt
+++ b/source/reference/api/reverse.txt
@@ -43,7 +43,7 @@ To use the ``reverse`` endpoint:
 - ``mongosync`` must be in the ``COMMITTED`` state.
 - Any indexes in the cluster must be :ref:`unique indexes
   <index-type-unique>`.You can convert a non-unique index to a unique
-  index, with the :dbcommand:`collMod` command. See `this example
+  index with the :dbcommand:`collMod` command. See `this example
   <https://www.mongodb.com/docs/v6.0/reference/command/collMod/#convert-an-existing-index-to-a-unique-index>`__
   for more details.
 
--- a/source/reference/api/reverse.txt
+++ b/source/reference/api/reverse.txt
@@ -42,7 +42,7 @@ To use the ``reverse`` endpoint:
   ``true``.
 - ``mongosync`` must be in the ``COMMITTED`` state.
 - Any indexes in the cluster must be :ref:`unique indexes
-  <index-type-unique>`.You can =convert a non-unique index to a unique
+  <index-type-unique>`.You can convert a non-unique index to a unique
   index, with the :dbcommand:`collMod` command. See `this example
   <https://www.mongodb.com/docs/v6.0/reference/command/collMod/#convert-an-existing-index-to-a-unique-index>`__
   for more details.


source/reference/api/reverse.txt
```diff
@@ -40,8 +40,12 @@ To use the ``reverse`` endpoint:
 - You must have started your original sync operation with the
   ``reversible`` and ``enableUserWriteBlocking`` options set to
   ``true``.
-
 - ``mongosync`` must be in the ``COMMITTED`` state.
+- Any indexes in the cluster must be :ref:`unique indexes
+  <index-type-unique>`.You can convert a non-unique index to a unique
```
mdb-ashley, line 45: [typo] Is the space between the two sentences missing?

mdb-ashley (approved):
LGTM with typo fix! Thanks so much, Dave.

davemungo (commented):
Thanks for the feedback @mdb-ashley 

Sending on to external review next

davemungo:
@ldennis this one is a holdover from the MDB World doc. When you have a moment would you have a look?

ldennis (changes requested):
This is not the prerequisite we wanted to add. We support non-unique indexes for /reserve as well. But for unique indexes, as part of /reverse, mongosync will need to convert those unique indexes to non-unique in order to support parallel replication. However, only unique indexes in the new format (built with binary >= 4.2) can be converted to non-unique. Therefore, the prerequisite of using /reverse is that users should make sure their unique indexes are in the appropriate format. To do this, we will require an initial sync for all nodes on the original source that have indexes created before 4.2. This will rebuild their indexes in the correct format.

For more details, please see [this doc](https://docs.google.com/document/d/1kgroil2tZ7Ivx8xbSfyZ2rEadbCndJ7D6CF3S4YWTNU/edit?usp=sharing). And I am happy to chat if you have other questions.

source/reference/api/reverse.txt
```diff
@@ -40,8 +40,14 @@ To use the ``reverse`` endpoint:
...
   ``reversible`` and ``enableUserWriteBlocking`` options set to
   ``true``.
-
 - ``mongosync`` must be in the ``COMMITTED`` state.
+- :ref:`Unique indexes <index-type-unique>` must be formatted properly.
+  Format indexes with the :dbcommand:`collMod` command. See `this example
+  <https://www.mongodb.com/docs/v6.0/reference/command/collMod/#convert-an-existing-index-to-a-unique-index>`__
+  for more details.
+- Source and destination clusters must be MongoDB 6.0 or later. If a
+  cluster has been upgraded from MongoDB 4.2 or earlier, an initial sync
```
davemungo, line 49: @ldennis What is an initial sync in this context? Is it a forward sync from the legacy 4.2 cluster to the other cluster before reversing? 

What happens if both clusters were upgraded for 4.2?

ldennis: Initial sync means https://www.mongodb.com/docs/manual/tutorial/resync-replica-set-member/#resync-a-member-of-a-replica-set

So they will have to resync a mongod, I think we should link that.

ldennis: They need to initial sync _all_ of their nodes.

source/reference/api/reverse.txt
```diff
@@ -41,11 +41,13 @@ To use the ``reverse`` endpoint:
...
 - ``mongosync`` must be in the ``COMMITTED`` state.
-- Any indexes in the cluster must be :ref:`unique indexes
-  <index-type-unique>`. You can convert a non-unique index to a unique
-  index with the :dbcommand:`collMod` command. See `this example
+- :ref:`Unique indexes <index-type-unique>` must be formatted properly.
+  Format indexes with the :dbcommand:`collMod` command. See `this example
   <https://www.mongodb.com/docs/v6.0/reference/command/collMod/#convert-an-existing-index-to-a-unique-index>`__
   for more details.
+- Source and destination clusters must be MongoDB 6.0 or later. If a
+  cluster has been upgraded from MongoDB 4.2 or earlier, an initial sync
```
ldennis, line 49: Technically, they only need to re-sync their nodes if they have unique indexes built with 4.2 and below.

davemungo (commented):
@ldennis Thanks for your feedback

source/reference/api/reverse.txt
```diff
@@ -40,8 +40,15 @@ To use the ``reverse`` endpoint:
 - You must have started your original sync operation with the
   ``reversible`` and ``enableUserWriteBlocking`` options set to
   ``true``.
-
 - ``mongosync`` must be in the ``COMMITTED`` state.
+- :ref:`Unique indexes <index-type-unique>` must be formatted properly.
+  Format indexes with the :dbcommand:`collMod` command. See `this example
+  <https://www.mongodb.com/docs/v6.0/reference/command/collMod/#convert-an-existing-index-to-a-unique-index>`__
+  for more details.
```
ldennis, line 47: I don't think this is relevant here. On reverse, mongosync converts the unique indexes to non-unique and that's why the unique indexes format matters here.

I think it would be more straightforward to say: "To use reverse replication, [unique indexes](https://www.mongodb.com/docs/master/core/index-unique/?&_ga=2.162030843.612489097.1658153614-1269726939.1563474234#std-label-index-type-unique) on the original source cluster must be formatted properly in the newer internal format. Source and destination clusters must also be MongoDB 6.0 or later. If an upgraded cluster has unique indexes that were created in MongoDB 4.2 or earlier, you must [resync](https://www.mongodb.com/docs/v6.0/tutorial/resync-replica-set-member/) all nodes in your cluster"

davemungo (commented):
@ldennis  thanks for the feedback, 

PR #90, opened 2022-08-11
DOCSP-24022-connection-string-mongosync

jason-price-mongodb:
JIRA

https://jira.mongodb.org/browse/DOCSP-24022

STAGE

https://docs-mongodbcom-staging.corp.mongodb.com/cluster-sync/docsworker-xlarge/DOCSP-24022-connection-string-mongosync/connecting/atlas-to-atlas/#connect-the-source-and-destination-clusters-with-mongosync

### DIFF
--- a/source/includes/example-connect.rst
+++ b/source/includes/example-connect.rst
@@ -50,18 +50,18 @@ following command on one line:
                        clusterTwo02.fancyCorp.com:20020,
                        clusterTwo03.fancyCorp.com:20020
 
-Atlas clusters require TLS/SSL connections. To use ``mongosync`` with
-Atlas clusters, you add the :urioption:`ssl=true <ssl>` option. For
+Atlas clusters require TLS connections. To use ``mongosync`` with
+Atlas clusters, you add the :urioption:`tls=true <tls>` option. For
 example:
 
 .. code-block:: shell
 
    mongosync \
-      --cluster0 'mongodb://clusterAdmin:superSecret@clusterOne01.fancyCorp.com:20020,clusterOne02.fancyCorp.com:20020,clusterOne03.fancyCorp.com:20020/?ssl=true' \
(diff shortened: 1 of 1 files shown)


jmd-mongo (approved):
Thanks for the updates, @jason-price-mongodb! This one LGTM.

Thanks,
Joe

jason-price-mongodb:
Hey Nikhil, couldn't find you in the system, so cannot add you as a reviewer. Please review. Thanks, Jason.

nikhil-mongo:
Hey Jason, 

1)
- We don't need extra text block to show the `tls` option, it can be changed in L46-L51 and L59-R61 can be removed. Since this section of document focuses on Atlas, it is important to have one example and not many.
- The exact change required is `/admin?tls=true`. I can see only `/?tls=true` option. So please make this correction too.

2)
- The DOCSP mentions two documentation links which are impacted due to this. The second one is - https://www.mongodb.com/docs/cluster-to-cluster-sync/current/connecting/onprem-to-atlas/#connect-the-source-and-destination-clusters-with-mongosync. Therefore, a change there would also be required for the second part of the mongosync example.

Let me know if I have missed to notice something on how these docs work on the second part of my review.

jason-price-mongodb:
@nikhil-mongo Hey Nikhil, one thing to keep in mind is the examples also need to cover non-Atlas clusters because there might be a case where a local cluster and Atlas cluster are used at the same time, so those lines you mention for delete should stay in the content to ensure the reader sees all the different permutations. I've added ``admin`` to the example. Also, the second page you mention contains the updated example because the text file in the doc update is also referenced in that page so it gets the example. Please click the link you mention and you'll see the updated example. Cheers!

nikhil-mongo:
LGTM

jason-price-mongodb:
Per Nikhil's LGTM in the comments, will squash and merge. I couldn't add him as a reviewer in this repo.

PR #96, opened 2022-10-25
DOCSP-25767 add versioning documentation v1.0

davemungo:
[VERSIONING](https://docs-mongodbcom-staging.corp.mongodb.com/cluster-sync/docsworker-xlarge/DOCSP-25767-add-versioning-documentation-v1.0/reference/versioning/)
[INDEX PAGE](https://docs-mongodbcom-staging.corp.mongodb.com/cluster-sync/docsworker-xlarge/DOCSP-25767-add-versioning-documentation-v1.0/)
[LIMITATIONS](https://docs-mongodbcom-staging.corp.mongodb.com/cluster-sync/docsworker-xlarge/DOCSP-25767-add-versioning-documentation-v1.0/reference/limitations/)

[JIRA](https://jira.mongodb.org/browse/DOCSP-25767)

### DIFF
--- a/source/reference/limitations.txt
+++ b/source/reference/limitations.txt
@@ -54,11 +54,10 @@ General Limitations
 - `Queryable Encryption
   <https://www.mongodb.com/docs/v6.0/core/queryable-encryption/>`__ is
   not supported.
-- If you replace the ``mongosync`` binary during an upgrade or a
-  downgrade, when you start the replacement binaries the ``mongosync``
-  processes will not resume any work that may have been in progress.
-  Syncing operations restart from the beginning when you start the new
-  processes.
+- After you replace the ``mongosync`` binary during an upgrade or a
+  downgrade, you should drop all non-system databases in the destination
+  cluster. Syncing operations will restart from the beginning when you
+  start the new processes.
 
 MongoDB Community Edition
 -------------------------
--- a/source/reference/limitations.txt
+++ b/source/reference/limitations.txt
@@ -57,7 +57,7 @@ General Limitations
 - If you change the ``mongosync`` binary during an upgrade or a
   downgrade, the replacement ``mongosync`` processes will not resume any
   work that may have been in progress. You must restart any prior
-  synching operations from the beginning.
+  syncing operations from the beginning.
 
 MongoDB Community Edition
 -------------------------
--- a/source/reference/limitations.txt
+++ b/source/reference/limitations.txt
@@ -56,8 +56,8 @@ General Limitations
   not supported.
 - After you replace the ``mongosync`` binary during an upgrade or a
   downgrade, you should drop all non-system databases in the destination
-  cluster. Syncing operations will restart from the beginning when you
-  start the new processes.
+  cluster before starting the new binary. Syncing operations will
+  restart from the beginning when you start the new processes.
 
 MongoDB Community Edition
 -------------------------
--- a/source/reference/limitations.txt
+++ b/source/reference/limitations.txt
@@ -54,6 +54,10 @@ General Limitations
 - `Queryable Encryption
   <https://www.mongodb.com/docs/v6.0/core/queryable-encryption/>`__ is
   not supported.
+- If you change the ``mongosync`` binary during an upgrade or a
+  downgrade, the replacement ``mongosync`` processes will not resume any
+  work that may have been in progress. You must restart any prior
+  synching operations from the beginning.
 
 MongoDB Community Edition
 -------------------------
--- a/source/reference/limitations.txt
+++ b/source/reference/limitations.txt
@@ -54,10 +54,11 @@ General Limitations
 - `Queryable Encryption
   <https://www.mongodb.com/docs/v6.0/core/queryable-encryption/>`__ is
   not supported.
-- If you change the ``mongosync`` binary during an upgrade or a
-  downgrade, the replacement ``mongosync`` processes will not resume any
-  work that may have been in progress. You must restart any prior
-  syncing operations from the beginning.
+- If you replace the ``mongosync`` binary during an upgrade or a
+  downgrade, when you start the replacement binaries the ``mongosync``
+  processes will not resume any work that may have been in progress.
+  Syncing operations restart from the beginning when you start the new
+  processes.
 
 MongoDB Community Edition
 -------------------------
--- a/source/reference/versioning.txt
+++ b/source/reference/versioning.txt
@@ -130,19 +130,20 @@ Major Releases
   destination cluster for a given set of input and configuration
   details. Except:
 
-  - Bug fixes for cases where ``mongosync`` failed to replicate data.
+  - Bug fixes for cases where ``mongosync`` fails to replicate data.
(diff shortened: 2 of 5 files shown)


jmd-mongo (approved):
Thanks for the updates, @davemungo. This LGTM mod a small terminology nit.

Thanks!
Joe

davemungo:
@autarch @tfogo   Please review when you have a moment.  I didn't add Ryan since Slack says he's on leave. 

The update drops most of the detailed examples of version changes since they seem more suited for internal use. 

The update also fixes two related points that were raised in the doc comments.

source/reference/limitations.txt
```diff
@@ -54,6 +54,10 @@ General Limitations
 - `Queryable Encryption
   <https://www.mongodb.com/docs/v6.0/core/queryable-encryption/>`__ is
   not supported.
+- If you change the ``mongosync`` binary during an upgrade or a
```
autarch, line 57: I think this is a little unclear. Maybe say:

> If you replace the \`\`mongosync\`\` binary in order to upgrade or downgrade it, when you restart the replacement ...

source/reference/versioning.txt
```diff
@@ -0,0 +1,120 @@
...
+Only the latest version in each major release series receives new patch
+releases. For example, when version {+c2c-product-name+} 2.1.0 is
+released, version 2.0.x would no longer receive patch releases. At the
+same time, version 1.3.2 would continue to receive patches until version
+1.4.0 was released.
+
+Upgrade and Downgrade
+~~~~~~~~~~~~~~~~~~~~~
+
+To upgrade, or downgrade, ``mongosync``:
```
autarch, line 88: nit: I think these commas are needed. I'm not sure if they're _wrong_ but it reads oddly to me with the commas present.

source/reference/versioning.txt
```diff
@@ -0,0 +1,120 @@
...
+- Stop all currently running ``mongosync`` processes.
+- Drop all non-system databases in the destination cluster.
+- :ref:`Install <c2c-install>` new ``mongosync`` binaries.
+- :ref:`Start <c2c-api-start>` the ``mongosync`` processes using the new
+  binaries.
+
+.. warning::
+ 
+   The new ``mongosync`` processes do not resume any work that may have
+   been in progress. You must restart any ongoing syncing operations
```
autarch, line 99: Same issue with "You must restart" as earlier.

source/reference/limitations.txt
```diff
@@ -54,6 +54,10 @@ General Limitations
...
+  downgrade, the replacement ``mongosync`` processes will not resume any
+  work that may have been in progress. You must restart any prior
```
autarch, line 59: Saying "You must restart" suggests that the user has to do something special to make this happen. But that's not the case. Syncing _will_ be restarted from the beginning simply by starting the sync process.

source/reference/versioning.txt
```diff
@@ -0,0 +1,120 @@
...
+   from the beginning.
+
+
+Persistent Metadata
+~~~~~~~~~~~~~~~~~~~
+
+During normal operation, ``mongosync`` creates metadata that is
+persisted to disk. This metadata is not versioned and may change at any
```
autarch, line 107: Maybe say "to disk in the destination database" to make it clear that it's not stored locally as files on the system running `mongosync`.

autarch (changes requested):
Overall, this looks good to me. I just had a few small comments.

davemungo (commented):
@autarch thanks for your feedback. This is ready for another pass

source/reference/limitations.txt
```diff
@@ -54,6 +54,11 @@ General Limitations
 - `Queryable Encryption
   <https://www.mongodb.com/docs/v6.0/core/queryable-encryption/>`__ is
   not supported.
+- If you replace the ``mongosync`` binary during an upgrade or a
```
autarch, line 57: This isn't quite right. It _will try_ to resume from where it left off. But we don't guarantee that will work. That's because we store in the destination cluster about the state of the sync process. We don't guarantee that this data will be compatible with other mongosync versions. So it's possible that it will either A) just work fine and resume where you left off; B) crash; C) do something else surprising.

Right now we want to tell people to drop the existing sync data (or just wipe the destination cluster entirely) as part of the upgrade/downgrade process. I see that we _do_ say this in the upgrade/downgrade section, but should we spell that out here as well?

autarch (changes requested):
One last thing, otherwise LGTM!

davemungo (commented):
@autarch  Thanks for the feedback, I've incorporated the suggestion.

autarch (approved):
LGTM! Thanks!

source/reference/limitations.txt
```diff
@@ -54,6 +54,10 @@ General Limitations
 - `Queryable Encryption
   <https://www.mongodb.com/docs/v6.0/core/queryable-encryption/>`__ is
   not supported.
+- After you replace the ``mongosync`` binary during an upgrade or a
+  downgrade, you should drop all non-system databases in the destination
+  cluster. Syncing operations will restart from the beginning when you
```
tfogo, line 59: Can we add "before starting the new binary" to this sentence? I think we should be clear that dropping must happen before the new binary is started.

source/reference/versioning.txt
```diff
@@ -0,0 +1,119 @@
+.. _c2c-release-version-numbers:
+
+===============================
+``mongosync`` Version Numbering
```
tfogo, line 4: This is a small thing, but "Version Numbering" seems far too narrow. I think we should just call this "Versioning" since the way the binary is versioned is more than just the numbering. For reference, the server also calls their equivalent page "Versioning": https://www.mongodb.com/docs/manual/reference/versioning/

source/reference/versioning.txt
```diff
@@ -0,0 +1,119 @@
...
+- If a version of {+c2c-product-name+} only works with an unsupported 
+  version of MongoDB Server, that version of {+c2c-product-name+} is
+  also unsupported.
+
+Patch Releases
+~~~~~~~~~~~~~~
+
+Only the latest version in each major release series receives new patch
+releases. For example, when version {+c2c-product-name+} 2.1.0 is
+released, version 2.0.x would no longer receive patch releases. At the
```
tfogo, line 81: It is a bit weird to me that we say 2.0.x or 1.3.2 would receive patches. I think it would be more idiomatic to say that 2.0 or 1.3 would receive patches. i.e. just reference the minor release.

tfogo (changes requested):
Looking great, thanks! I have a few suggestions.

Also, it looks like we omitted the "Details on Major, Minor, and Patch Release Changes" section from the policy doc Dave R drafted. Could we include that? We have a summary of when major, minor, and patch numbers will be incremented, but I think it would be nice to publicly document the details.

davemungo (commented):
@tfogo I've updated the text and added an examples section as discussed in slack. Would you have a look?

tfogo (approved):
Sorry for the delay with this. LGTM!

PR #112, opened 2023-02-17
DOCSP-28129 error response

davemungo:
[STAGING](https://docs-mongodbcom-staging.corp.mongodb.com/cluster-sync/docsworker-xlarge/DOCSP-28129-progress-endoint-v1.2/reference/api/progress/)

[JIRA](https://jira.mongodb.org/browse/DOCSP-28129)

[build](https://workerpool-boxgs.mongodbstitch.com/pages/job.html?collName=queue&jobId=63efc14b05dd3cc814e65aec)

### DIFF
--- a/source/includes/api/tables/basic-response.rst
+++ b/source/includes/api/tables/basic-response.rst
@@ -13,10 +13,8 @@
 
    * - ``error``
      - string
-     - If an error occurred, indicates the name of the error. This field
-       is omitted from the response when ``success`` is ``true``.
+     - .. include:: /includes/api/tables/text-error.rst
 
    * - ``errorDescription``
      - string
-     - Detailed description of the error that occurred. This field is
-       omitted from the response when ``success`` is ``true``.
+     - .. include:: /includes/api/tables/text-errorDescription.rst
--- a/source/includes/api/tables/basic-response.rst
+++ b/source/includes/api/tables/basic-response.rst
@@ -11,6 +11,6 @@
      - boolean
      - When the request is successful, this value is ``true``.
 
-   .. include:: /includes/api/tables/response-errorDescription.rst
-    
-   .. include:: /includes/api/tables/response-errorDescription.rst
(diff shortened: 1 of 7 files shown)


jmd-mongo (approved):
This one LGTM! Thanks for the updates, @davemungo.

davemungo:
@autarch would you have a look at this docs PR when you get a moment?

autarch (changes requested):
This works the same way for all the other endpoints, so I think this should be shown in the same way.

With `/progress`, you _either_ get the various info fields _or_ a document with just the error info. You can't get both, but the way this is formatted makes it look like you can.

davemungo (commented):
@autarch  I'm confused, the endpoints are all formatted the same way - all possible responses are listed in a single table. I don't think it makes sense to break all of the endpoints into two tables (error and success) since all of the tables except this one are only 3 lines long. 

Instead, I've changed the wording for the /progress endpoint and added an introductory line to try to highlight the distinction between status and error messages.

autarch:
> @autarch I'm confused, the endpoints are all formatted the same way - all possible responses are listed in a single table. I don't think it makes sense to break all of the endpoints into two tables (error and success) since all of the tables except this one are only 3 lines long.

Doh, _I_ was confused. Now I see you're correct. It's just that the other endpoints don't really have a significant body unless there's an error, so it looks like it's just documenting an error response.

> Instead, I've changed the wording for the /progress endpoint and added an introductory line to try to highlight the distinction between status and error messages.

This is helpful, I think.

Longer-term, I think it's on us to change how this behaves. We should not be returning a 200 status for errors, and we should document the 200 response separately from other statuses.

davemungo (commented):
@autarch   Sounds good.  Is this ok to merge in for now?

autarch (approved):
LGTM!

autarch:
> @autarch Sounds good. Is this ok to merge in for now?

Yes, sorry for the confusion.

PR #123, opened 2023-04-11
DOCSP-28806 mongosyncID/coordinatorID return fields for /progress calls

kennethdyer:
## Description

Adds the `mongosyncID` and `coordinatorID` fields to the `/progress` output table.

## Staging
[Response](https://docs-mongodbcom-staging.corp.mongodb.com/cluster-sync/docsworker-xlarge/DOCSP-28806-coordinatorid/reference/api/progress/#response) (last two items on the table).

## Jira

[DOCSP-28806](https://jira.mongodb.org/browse/DOCSP-28806)

## Build

[Build](https://workerpool-boxgs.mongodbstitch.com/pages/job.html?collName=queue&jobId=6435a05bb49106a49829dd29)

### DIFF
--- a/source/includes/api/tables/progress-response.rst
+++ b/source/includes/api/tables/progress-response.rst
@@ -84,14 +84,17 @@
    * - ``mongosyncID``
      - string
      - The identifier string for the ``mongosync`` instance.
-       To configure this value, see the :setting:`id` setting.
 
    * - ``coordinatorID``
      - string
-     - The identifier string for the coordinator instance.  To change this value,
-       see the :setting:`id` setting on the coordinator instance of ``mongosync``.
+     - The identifier string for the coordinator instance.  
 
-       When the ``mongosync`` instance is the coordinator, this field returns
-       ``null``.
+       - When ``mongosync`` is coordinated by another instance, this field shows
+         the identifier string for the coordinator instance.
 
+       - When ``mongosync`` is a coordinator or runs alone, this field returns
+         the same value as its ``mongosyncID`` field.
+
+       - When ``mongosync`` starts up, this field returns ``null`` until
+         ``mongosync`` identifies the coordinator.
 
--- a/source/includes/api/tables/progress-response.rst
+++ b/source/includes/api/tables/progress-response.rst
@@ -80,6 +80,7 @@
      - string
      - If an error occurred, gives a detailed description of the error.
        This field is omitted when the call to the endpoint is successful
+
    * - ``mongosyncID``
      - string
      - The identifier string for the ``mongosync`` instance.
--- a/source/includes/api/tables/progress-response.rst
+++ b/source/includes/api/tables/progress-response.rst
@@ -80,4 +80,17 @@
      - string
      - If an error occurred, gives a detailed description of the error.
        This field is omitted when the call to the endpoint is successful
+   * - ``mongosyncID``
+     - string
+     - The identifier string for the ``mongosync`` instance.
+       To configure this value, see the :setting:`id` setting.
+
+   * - ``coordinatorID``
+     - string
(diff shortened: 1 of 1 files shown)


ianf-mongodb (approved):
LGTM % 1 nit.

source/includes/api/tables/progress-response.rst
```diff
@@ -80,4 +80,17 @@
      - string
      - If an error occurred, gives a detailed description of the error.
        This field is omitted when the call to the endpoint is successful
+   * - ``mongosyncID``
```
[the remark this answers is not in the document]
kennethdyer: Fixed

source/includes/api/tables/progress-response.rst
```diff
@@ -81,3 +81,17 @@
...
 
+   * - ``mongosyncID``
+     - string
+     - The identifier string for the ``mongosync`` instance.
+       To configure this value, see the :setting:`id` setting.
+
+   * - ``coordinatorID``
+     - string
+     - The identifier string for the coordinator instance.  To change this value,
+       see the :setting:`id` setting on the coordinator instance of ``mongosync``.
```
suspicious-mango, line 92: I am not sure what the behavior would be if the `id` of the coordinator instance is changed. Assuming that a restart is needed, would the resulting mongosync still be the coordinator instance?

Please check in #mongosync-impl, but in general, I think we shouldn't tell users about the possibility of changing `coordinatorID`. Changing `mongosyncID` is still permitted, and the side-effect of that concerning which instance is the coordinator should be documented there.

source/includes/api/tables/progress-response.rst
```diff
@@ -81,3 +81,17 @@
...
+
+       When the ``mongosync`` instance is the coordinator, this field returns
+       ``null``.
```
suspicious-mango, line 95: This is incorrect, when the mongosync instance is the coordinator, the values of mongosyncID and coordinatorID will be equal.

coordinatorID could be null during mongosync start-up.

suspicious-mango (changes requested):
please check in #mongosync-impl about the behavior of mongosync if `:setting:\`id\`` is used to modify mongosyncID.

source/includes/api/tables/progress-response.rst
edobranov: This isn't exactly possible / how it works. There's nothing tying a mongosync kernel process / PID to the coordinator role.

The only persistent notion of a coordinator role is what the value is of the `coordinator` field in the destination metadata (specifically the global state document). This is derived from one of the mongosyncs' `--id` CLI value and stays constant throughout a migration and "cannot" be changed. If you wanted to change this, you'd have to manually change the `coordinator` field in a database that's called `mongosync_reserved_for_internal_use`, but we broadly say that any manual adjustments to such metadata will lead to undefined behavior.

If, say, a user were running two mongosyncs (each on a different machine) and wanted switch which mongosync acts as the coordinator because it's on a preferred machine, they'd have to shut down both processes and swap their `--id` values on the command line. This basically switches their "roles".

Still, I agree that we probably shouldn't be advertising this to users right now. We'd probably need separate, more formal documentation around this, since it's not totally obvious how this works.

kennethdyer: Fixed

source/includes/api/tables/progress-response.rst
kennethdyer: Fixed

suspicious-mango (approved):

kennethdyer:
@suspicious-mango , changes applied, anything else you'd like to see?

suspicious-mango:
@kennethdyer LGTM.

PR #129, opened 2023-06-13
DOCSP-30223 1.4.0 release notes

davemungo:
[Rel notes](https://docs-mongodbcom-staging.corp.mongodb.com/cluster-sync/docsworker-xlarge/DOCSP-30223-1.4.0-release-notes/release-notes/1.4/)

[JIRA](https://jira.mongodb.org/browse/DOCSP-30223)

[BUILD](https://workerpool-boxgs.mongodbstitch.com/pages/job.html?collName=queue&jobId=6489d28a67a273b81629ba14)  Updated 2023-06-14

### DIFF
--- a/source/reference.txt
+++ b/source/reference.txt
@@ -17,5 +17,6 @@ Reference
    /reference/limitations
    /reference/logging
    /reference/permissions
+   /reference/telemetry
    /reference/versioning
 
--- a/source/release-notes/1.4.txt
+++ b/source/release-notes/1.4.txt
@@ -37,6 +37,13 @@ Faster Processing for Sharded Clusters
 Version 1.4.0 improves performance for replica sets and single-shard
 clusters. Multi-shard clusters have even better speed improvements.
 
+Improved Memory Handling
+~~~~~~~~~~~~~~~~~~~~~~~~
+
+In version 1.4.0, ``mongosync`` uses less memory to buffer change
+events. Improved memory handling alleviates some out-of-memory (OOM)
+issues.
+
 Telemetry
 ~~~~~~~~~
 
--- a/source/release-notes/1.4.txt
+++ b/source/release-notes/1.4.txt
@@ -29,8 +29,7 @@ Improved Resiliency
 ~~~~~~~~~~~~~~~~~~~
 
 ``mongosync`` 1.4.0 improves :term:`oplog` handling. The ``oplog``
-handling improvement adds resiliency to migrations and reduces the time
-to complete large migrations.
+handling improvement adds resiliency to long running migrations.
 
 Faster Processing for Sharded Clusters
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
@@ -48,19 +47,16 @@ data to improve MongoDB products. You can disable collection of this
 Bug Fixes
 ----------
 
-- ``REP-2148`` Fixes an issue so documents with ``_id`` values that
-  evaluate to the same string replicate correctly.
+- Fixes an issue so documents with ``_id`` values that evaluate to the
+  same string replicate correctly.
 
-- ``REP-2411`` Fixes a collection naming issue. Prior to version 1.4.0,
-  ``mongosync`` crashes when a new collection reuses a dropped
-  collection's name with letter case changes.
+- Fixes a collection naming issue. Prior to version 1.4.0, ``mongosync``
+  crashes when a new collection reuses a dropped collection's name with
+  letter case changes.
   
   For example, previous versions of ``mongosync`` crash if a new
   ``SampleName`` collection replaces the dropped ``samplename``
   collection.
-  
-- ``REP-2378`` Fixes an issue caused by a server update,
-  ``BACKPORT-15099``.
 
 Additional Considerations
 -------------------------
--- a/source/release-notes/1.4.txt
+++ b/source/release-notes/1.4.txt
@@ -31,11 +31,11 @@ Improved Resiliency
 ``mongosync`` 1.4.0 improves :term:`oplog` handling. The ``oplog``
 handling improvement adds resiliency to long running migrations.
 
-Faster Processing for Sharded Clusters
-~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+Faster Processing
+~~~~~~~~~~~~~~~~~
 
-Version 1.4.0 improves performance for replica sets and single-shard
-clusters. Multi-shard clusters have even better speed improvements.
+Version 1.4.0 improves ``change event application`` speeds for improved
+performance.
 
 Improved Memory Handling
 ~~~~~~~~~~~~~~~~~~~~~~~~
--- a/source/release-notes/1.4.txt
+++ b/source/release-notes/1.4.txt
@@ -64,8 +64,8 @@ Additional Considerations
 The MongoDB server's :dbcommand:`validate` command may report data
 corruption on destination clusters that run MongoDB version 6.0 through
 6.0.6. The data on the destination server is not corrupt. A bug in the
-:ref:`WiredTiger <storage-wiredtiger>` database engine, ``WT-11051``,
-causes the false report.
+:ref:`WiredTiger <storage-wiredtiger>` database engine,
+:issue:`WT-11051`, causes the false report.
 
 Minimum Supported Version
 -------------------------
--- a/source/release-notes/1.4.txt
+++ b/source/release-notes/1.4.txt
@@ -12,9 +12,7 @@ Release Notes for ``mongosync`` 1.4
    :depth: 2
    :class: singlecol
 
-.. note::
-
-   ``mongosync`` 1.4 Released <<TBD>>, 2023
+**Version 1.4 released June 16, 2023**
 
 New Features
 -------------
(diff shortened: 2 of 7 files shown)


davemungo (commented):
Notes for reviewers

source/reference.txt
```diff
@@ -17,5 +17,6 @@ Reference
    /reference/limitations
    /reference/logging
    /reference/permissions
+   /reference/telemetry
```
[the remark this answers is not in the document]
mdb-ashley: [praise] Nice catch!

mdb-ashley (approved):
LGTM! Nice work on this, Dave.

davemungo:
@alan-zheng-mdb  Would you review the 1.4.0 C2C release notes?  Thanks!

source/release-notes/1.4.txt
```diff
@@ -27,7 +25,54 @@ Starting in 1.4.0, you can set the workload level that
...
+~~~~~~~~~
+
+Starting in 1.4.0, ``mongosync`` collects anonymous, aggregated usage
+data to improve MongoDB products. You can disable collection of this
+:ref:`telemetry <c2c-telemetry>` data when you start ``mongosync``.
+
+Bug Fixes
+----------
+
+- ``REP-2148`` Fixes an issue so documents with ``_id`` values that
```
mmcclimon, line 51: I think we should get rid of the JIRA ticket IDs here. None of the other mongosync release notes have them, and the REP project isn't publicly accessible, so people can't actually use them to get more information.

source/release-notes/1.4.txt
```diff
@@ -27,7 +25,54 @@ Starting in 1.4.0, you can set the workload level that
...
+- ``REP-2411`` Fixes a collection naming issue. Prior to version 1.4.0,
+  ``mongosync`` crashes when a new collection reuses a dropped
+  collection's name with letter case changes.
+  
+  For example, previous versions of ``mongosync`` crash if a new
+  ``SampleName`` collection replaces the dropped ``samplename``
+  collection.
+  
+- ``REP-2378`` Fixes an issue caused by a server update,
+  ``BACKPORT-15099``.
```
mmcclimon, line 63: We can drop this bullet entirely. This ticket only affects version 4.4 anyway, which we don't publicly support. (Sorry about that; I'll make a note that this document should differentiate between public and private support in the future!)

source/release-notes/1.4.txt
```diff
@@ -27,7 +25,54 @@ Starting in 1.4.0, you can set the workload level that
 destination clusters. For more information, see the
 :option:`--loadLevel` option.
 
+Improved Resiliency
+~~~~~~~~~~~~~~~~~~~
+
+``mongosync`` 1.4.0 improves :term:`oplog` handling. The ``oplog``
+handling improvement adds resiliency to migrations and reduces the time
+to complete large migrations.
```
mmcclimon, line 33: I find this a little confusing. The thing we're now more resilient is specifically "oplog rollover". The problem with older versions is that if the first part of a migration took (say) 48 hours, and we only had 24 hours of oplog entries available, mongosync could not complete the migration. Now, we do more work in the first part of the migration so that there's less to do in the second part.

I'm not sure how best to describe that, but I'm tagging @edobranov here, who might have additional thoughts.

source/release-notes/1.4.txt
```diff
@@ -27,7 +25,54 @@ Starting in 1.4.0, you can set the workload level that
...
+- ``REP-2378`` Fixes an issue caused by a server update,
+  ``BACKPORT-15099``.
+
+Additional Considerations
+-------------------------
+
+The MongoDB server's :dbcommand:`validate` command may report data
+corruption on destination clusters that run MongoDB version 6.0 through
+6.0.6. The data on the destination server is not corrupt. A bug in the
+:ref:`WiredTiger <storage-wiredtiger>` database engine, ``WT-11051``,
```
mmcclimon, line 71: (To be clear: I think this Jira ID can stay, because the WT project is publicly accessible. It might be useful to link it to https://jira.mongodb.org/browse/WT-11051)

mmcclimon (changes requested):
Looks pretty good, thanks! I have made a few suggestions and asked Evgeni to chime in to help with the Oplog Rollover Resilience bullet point.

source/release-notes/1.4.txt
davemungo: @mmcclimon I added them in for two reasons 

- One is because we have the issue numbers on the server release notes so this way matches the server approach. Those are public though so maybe they are more helpful there. 

- The other reason is more public facing. I thought perhaps it would be useful for customers/support to be able to refer to particular issues. 


No problem dropping them if you think that's better.

source/release-notes/1.4.txt
davemungo: Done

source/release-notes/1.4.txt
davemungo: OK

davemungo (commented):
@mmcclimon  thanks for your feedback, this is ready for another look

source/release-notes/1.4.txt
```diff
@@ -27,7 +25,50 @@ Starting in 1.4.0, you can set the workload level that
...
+
+Additional Considerations
+-------------------------
+
+The MongoDB server's :dbcommand:`validate` command may report data
+corruption on destination clusters that run MongoDB version 6.0 through
+6.0.6. The data on the destination server is not corrupt. A bug in the
+:ref:`WiredTiger <storage-wiredtiger>` database engine,
+:issue:`WT-11051`, causes the false report.
+
```
edobranov, line 69: Apologies for the late EOD comment -- I just wrapped up [REP-2770](https://jira.mongodb.org/browse/REP-2770) and think we may want to add something in here for that. The ticket's Release Notes describe the exact change, but we can also keep it high-level and say something like:

> Mongosync uses less memory for buffering change events, alleviating some out-of-memory issues.

(@mmcclimon if you agree / have thoughts)

mmcclimon: Yeah, I agree it would be good to put something here; it's a change end users will definitely notice!

edobranov (approved):

source/release-notes/1.4.txt
```diff
@@ -27,7 +25,57 @@ Starting in 1.4.0, you can set the workload level that
...
+~~~~~~~~~~~~~~~~~~~
+
+``mongosync`` 1.4.0 improves :term:`oplog` handling. The ``oplog``
+handling improvement adds resiliency to long running migrations.
+
+Faster Processing for Sharded Clusters
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Version 1.4.0 improves performance for replica sets and single-shard
+clusters. Multi-shard clusters have even better speed improvements.
```
alan-zheng-mdb, line 38: Since we don't quantify speed, should we combine and say replica sets and sharded clusters have improved performance? 

Is a particular stage (collection copy or CEA) sped up? @davemungo

mmcclimon: (The optimization work focuses on CEA.)

alan-zheng-mdb (approved):
lgtm

source/release-notes/1.4.txt
```diff
@@ -27,7 +25,57 @@ Starting in 1.4.0, you can set the workload level that
 destination clusters. For more information, see the
 :option:`--loadLevel` option.
 
+Improved Resiliency
+~~~~~~~~~~~~~~~~~~~
+
+``mongosync`` 1.4.0 improves :term:`oplog` handling. The ``oplog``
+handling improvement adds resiliency to long running migrations.
```
edobranov, line 32: @davemungo We had a close-out review meeting for this project earlier today and this point came up (apologies for not getting around to replying to the now-outdated comment on this).

I think we can accentuate the problem this is meant to address and provide a high-level overview of the mechanism, since it's one of the largest pain points customers currently experience and TSEs be curious about how this works. Something like this seems reasonable to @tfogo and myself:

> Mongosync 1.4.0 adds resiliency to oplog rollover.
> 
> Previously, mongosync waited for all of the initial data to be copied before it replicated any changes to that data. For a migration with a large initial data set, this wait period made mongosync vulnerable to losing its position in the oplog, requiring the migration to be restarted from scratch.
> 
> Now, mongosync can simultaneously copy the initial data set and apply changes to that data. This means that mongosync can maintain a recent position in the oplog and that the risk of migrations having to be restarted from scratch due to large initial data sets is greatly mitigated.
> 
> See [oplog sizing](https://www.mongodb.com/docs/cluster-to-cluster-sync/current/reference/oplog-sizing/) for details on monitoring your migration's risk of oplog rollover.

(we're planning to update that documentation linked at the end)

Happy to go over if you think this level of detail is suitable, needs adjustments, etc.

davemungo: Updated. 

There are some concepts here that we don't really explain in much detail (eg 'initial data sync', the oplog). I added a brief intro sentence for now, do we need to expand the concept pages after the release?

I removed references to 'migration' since I think this applies to the sync use case too. Let me know if I should revert to 'migration'.

I changed the wording a little to better match our style guide. Let me know if I inadvertently changed the substance.

davemungo (commented):
@tfogo, @edobranov  thanks for your feedback, this is ready for another look

PR #130, opened 2023-06-15
DOCSP-29500 Clarify oplog sizing

tfogo:
This is an example of the changes needed for DOCSP-29500. Key points:

- No longer say users need an oplog window large enough for the entirety of initial sync.
- No longer say that the size of the data set determines how large the oplog should be.
- Explain how mongosync operation application may be slower during initial sync and may speed up after initial sync
- Explain how users should check the lagTimesSeconds in the /progress endpoint and compare this with the oplog window to decide if they need to increase the oplog window.

I just wanted to get the main changes down on paper, I'm sure the language and formatting could be improved here. Please feel free to start a new PR. Just wanted to provide this as reference.

### DIFF
--- a/source/faq.txt
+++ b/source/faq.txt
@@ -33,18 +33,6 @@ Should I increase the size of the ``oplog`` in the source cluster?
 
 .. include:: /includes/fact-oplog-background
 
-The proper ``oplog`` size depends on system hardware, network speed,
-and other factors including system workload. However, assuming network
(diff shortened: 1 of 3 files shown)


davemungo:
Moved content from GitHub to

https://jira.mongodb.org/browse/DOCSP-30736 
and
https://jira.mongodb.org/browse/DOCSP-29500

Closing this out

PR #146, opened 2023-08-16
DOCSP-32048 process Error Response

kennethdyer:
## Description

Fix the `progress` documentation to indicate that the error response is different from a successful completion.  Also,
modifies the response table to allow for better wrapping.

## Staging

* [progress response](https://docs-mongodbcom-staging.corp.mongodb.com/cluster-sync/docsworker-xlarge/DOCSP-32048-process-clarification/reference/api/progress/#response)

## Jira

* [DOCSP-32048](https://jira.mongodb.org/browse/DOCSP-32048)

## Build

* [Build](https://workerpool-boxgs.mongodbstitch.com/pages/job.html?collName=queue&jobId=64dcf85f67dbe910102ca0f8)
* [Build](https://workerpool-boxgs.mongodbstitch.com/pages/job.html?collName=queue&jobId=64e4da3757e37806c5411521)

### DIFF
--- a/source/reference/api/progress.txt
+++ b/source/reference/api/progress.txt
@@ -44,6 +44,7 @@ Error Response
 If ``mongosync`` encounters an error, the ``progress`` endpoint instead
 returns an ``error`` object with the following field:
 
+.. include:: /includes/api/tables/progress-error-response.rst
 
 Behavior
 --------
--- a/source/reference/api/progress.txt
+++ b/source/reference/api/progress.txt
@@ -28,10 +28,17 @@ Request
 Response
 --------
 
-All response fields are wrapped in a top-level ``progress`` object. The
-endpoint returns either an updated status or an error.
+The ``progress`` endpoint returns etiher an updated status or an error.
+
+On a successful return, all response fields are wrapped in a top-level
+``progress`` object with the following fields:
 
 .. include:: /includes/api/tables/progress-response.rst
+
+If ``mongosync`` encounters an error, the ``progress`` endpoint instead
+returns an ``error`` object with the following field:
+
+.. include:: /includes/api/tables/progress-error-response.rst
 
 Behavior
 --------
--- a/source/reference/api/progress.txt
+++ b/source/reference/api/progress.txt
@@ -33,16 +33,17 @@ The ``progress`` endpoint returns etiher an updated status or an error.
 Successful Response
 ~~~~~~~~~~~~~~~~~~~
 
-On a successful return, all response fields are wrapped in a top-level
-``progress`` object with the following fields:
+If ``mongosync`` successfully gets the status of the sync process, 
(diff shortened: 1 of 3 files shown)


source/reference/api/progress.txt
```diff
@@ -28,11 +28,24 @@ Request
...
 --------
 
-All response fields are wrapped in a top-level ``progress`` object. The
-endpoint returns either an updated status or an error.
+The ``progress`` endpoint returns etiher an updated status or an error.
+
+Successful Response
+~~~~~~~~~~~~~~~~~~~
+
+On a successful return, all response fields are wrapped in a top-level
```
jeff-allen-mongo, line 36: [suggestion]

For parallelism with the Error Response heading, maybe we could say: "If ``mongosync`` successfuly gets the status of the synchronization process, ..."

source/reference/api/progress.txt
```diff
@@ -28,11 +28,24 @@ Request
...
+``progress`` object with the following fields:
 
 .. include:: /includes/api/tables/progress-response.rst
 
+Error Response
+~~~~~~~~~~~~~~
+
+If ``mongosync`` encounters an error, the ``progress`` endpoint instead
```
jeff-allen-mongo, line 44: [nit, minimalism]

I think we can remove "instead".

jeff-allen-mongo (approved):
LGTM % nits. Thanks @kennethdyer 🚀

source/reference/api/progress.txt
kennethdyer: Fixed

source/reference/api/progress.txt
kennethdyer: Fixed

source/reference/api/progress.txt
```diff
@@ -28,11 +28,25 @@ Request
...
 
 .. include:: /includes/api/tables/progress-response.rst
 
+Error Response
+~~~~~~~~~~~~~~
+
+If ``mongosync`` encounters an error, the ``progress`` endpoint returns 
+an ``error`` object with the following field:
+
+.. include:: /includes/api/tables/progress-error-response.rst
```
mmcclimon, line 48: This is not quite right. In the case of an error, the response is an object with only an error key, like this:

```
{
  "error": "Some description of the error"
}
```

There's no `errorDescription` field at all, and the object isn't nested. Sorry if this was unclear in the ticket!

mmcclimon (changes requested):
One thing to fix up; thanks!

source/reference/api/progress.txt
kennethdyer: Fixed

mmcclimon (approved):
Sorry I let this sit for so long! The PR process for the docs team isn't quite what we use internally, so this didn't come up in my normal view of things to review.

This LGTM, thanks so much!

PR #222, opened 2023-12-13
DOCSP 28174 adding info to commit page

ltran-mdb2:
## DESCRIPTION

adding information to the commit page to clarify behavior

## STAGING

https://preview-mongodbltranmdb2.gatsbyjs.io/cluster-sync/DOCSP-28174/reference/api/commit/

## JIRA

https://jira.mongodb.org/browse/DOCSP-28174

## BUILD LOG

https://workerpool-boxgs.mongodbstitch.com/pages/job.html?collName=queue&jobId=6579e76f7a5cb5fef5c7c6c4

## Self-Review Checklist

## External Review Requirements

### DIFF
--- a/source/reference/api/commit.txt
+++ b/source/reference/api/commit.txt
@@ -29,9 +29,8 @@ characteristics that ``mongosync`` alters during the synchronization process.
    * - Collection Characteristic
      - Impact of ``commit`` 
    * - :ref:`Unique Indexes <c2c-dr-unique>`
-     - ``commit`` resets unique indexes that
-       ``mongosync`` replicated as non-unique on the destination
-       cluster.
+     - ``commit`` resets unique indexes that``mongosync`` replicated as
+       non-unique on the destination cluster.
 
    * - :ref:`TTL Indexes <c2c-dr-ttl>`
      - ``commit`` resets ``expireAfterSeconds`` that ``mongosync`` set
@@ -42,13 +41,13 @@ characteristics that ``mongosync`` alters during the synchronization process.
        replicated as non-hidden on the destination cluster.
 
    * - :ref:`Write Blocking <c2c-dr-write-blocking>`
-     - if ``mongosync`` disabled writes on the destination cluster,
-     ``commit`` disables the Write Block Mode. 
+     - If ``mongosync`` disabled writes on the destination cluster,
+       ``commit`` disables the Write Block Mode. 
 
    * - :ref:`Capped Collections <c2c-dr-capped-collections>`
      - ``commit`` resets the required maximum size of capped collections
-        that ``mongosync`` set to the maximum allowable size on the
-        destination cluster.
+       that ``mongosync`` set to the maximum allowable size on the
+       destination cluster.
 
 Requirements
 ------------
--- a/source/reference/api/commit.txt
+++ b/source/reference/api/commit.txt
@@ -15,8 +15,7 @@
 Description
 -----------
 
-Commits the synchronization operation to the destination cluster.
-``commit`` finalizes the sync between the source cluster and the
+Finalizes the sync between the source cluster and the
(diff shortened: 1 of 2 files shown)


source/reference/api/commit.txt
```diff
@@ -16,6 +16,8 @@ Description
 -----------
 
 Commits the synchronization operation to the destination cluster.
```
ajhuh-mdb, line 18: [suggestion] It looks like the first half of the new sentence is saying virtually the same thing as the first sentence that's already here. Maybe we can just merge these two?
```suggestion
Finalizes the synchronization operation between the source and destination clusters. ``commit``stops continuous...
```

ajhuh-mdb (approved):
LGTM with one suggestion!

source/reference/api/commit.txt
ltran-mdb2: My only hesitation here is that the the other endpoints use the endpoint name to describe the definition of the endpoint behavior. I do like the idea of combining the sentences and using "finalize" rather than "commit" but just wanted to check in re consistency

ajhuh-mdb: Gotcha, that makes sense! I think the main concern is that the original JIRA ticket mentions that the "commit" language is what was causing confusion with the definition in the first place, so I think that could warrant overriding consistency in favor of clarity.

ltran-mdb2: Sounds good, just made the change!

FGasper (commented):
Everything here is correct.

I wasn’t sure if we want to explain more of what happens in a commit—e.g., indexes become unique-ified, collection caps are reduced to production levels, etc.—but TBH I don’t know how much of that we consider to be implementation.

I’ll ping the team.

ltran-mdb2:
Hi @FGasper ! I wanted to follow up on this and see if I should add anything else?

source/reference/api/commit.txt
```diff
@@ -15,7 +15,39 @@
...
+       replicated as non-hidden on the destination cluster.
+
+   * - :ref:`Write Blocking <c2c-dr-write-blocking>`
+     - If ``mongosync`` disabled writes on the destination cluster,
+       ``commit`` disables the Write Block Mode. 
+
+   * - :ref:`Capped Collections <c2c-dr-capped-collections>`
+     - ``commit`` resets the required maximum size of capped collections
+       which ``mongosync`` set to the maximum allowable size on the
+       destination cluster.
```
FGasper, line 50: Does this header duplicate the information in the disaster-recovery page?

If so, that seems like additional overhead and opportunity for confusion/inaccuracy, if those 2 pages now have to be kept manually in sync.

Is it possible to write the table out to a file, then include it separately? Maybe using rST’s [include](https://docutils.sourceforge.io/docs/ref/rst/directives.html#including-an-external-document-fragment) directive?

FGasper (changes requested):

source/reference/api/commit.txt
ltran-mdb2: The information in the commit table and the [disaster recovery table ](https://www.mongodb.com/docs/cluster-to-cluster-sync/current/reference/disaster-recovery/) is similar but has different use cases which is why I hesitate to make make them identical with an include file. For the disaster recovery page, the user needs to know which commands to use (for example, use the collMod command to set expireAfterSeconds to the desired value) to manually take action while the commit table does not need to provide that information. Because the disaster recovery table needs to include instructions on how to manually reset the changes as opposed to the commit table which is just describes how things happen automatically,  I would be in favor of leaving the two tables separate.

FGasper: That makes sense. Thanks!

FGasper (approved):
LGTM. Thanks!

PR #231, opened 2023-12-20
DOCSP 24715 mongosyncs must be the same version

ltran-mdb2:
## DESCRIPTION

clarifying that the same version of mongosync must be used between all instances

## STAGING

https://docs-mongodbcom-staging.corp.mongodb.com/cluster-sync/docsworker-xlarge/DOCSP-24715/multiple-mongosyncs/#configure-multiple-mongosync-instances

## JIRA

https://jira.mongodb.org/browse/DOCSP-24715

## BUILD LOG

https://workerpool-boxgs.mongodbstitch.com/pages/job.html?collName=queue&jobId=65831e407a5cb5fef5098dc8

## Self-Review Checklist

## External Review Requirements

### DIFF
--- a/source/multiple-mongosyncs.txt
+++ b/source/multiple-mongosyncs.txt
@@ -42,7 +42,7 @@ Configure Multiple ``mongosync`` Instances
 The number of ``mongosync`` instances must match the number of shards on
 the source cluster. For a replica set source, you can only use one
 ``mongosync`` instance. You must use the same version of ``mongosync``
-between instances. 
+between all instances. 
 
 To configure multiple ``mongosync`` instances:
 
--- a/source/multiple-mongosyncs.txt
+++ b/source/multiple-mongosyncs.txt
@@ -40,9 +40,9 @@ Configure Multiple ``mongosync`` Instances
 ------------------------------------------
 
 The number of ``mongosync`` instances must match the number of shards on
-the source cluster. For a replica set source, you can only use one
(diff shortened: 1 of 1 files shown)


source/multiple-mongosyncs.txt
```diff
@@ -41,7 +41,8 @@ Configure Multiple ``mongosync`` Instances
 
 The number of ``mongosync`` instances must match the number of shards on
 the source cluster. For a replica set source, you can only use one
-``mongosync`` instance. 
+``mongosync`` instance. You must use the same version of ``mongosync``
```
jeff-allen-mongo, line 44: [Question]

Would it make more sense to put this sentence after "The number of ``mongosync`` instances must match the number of shards on the source cluster." ?

This seems to be where we're talking about using multiple ``mongosync`` instances, so it might make sense to have this info there rather than where we're talking about only using one ``mongosync`` instance.

jeff-allen-mongo (commented):
Thanks for this update @ltran-mdb2! Just have a small question on the ordering of these sentences.

source/multiple-mongosyncs.txt
ltran-mdb2: That makes sense, just made the switch!

jeff-allen-mongo (approved):
LGTM!

ajayvijayakumar123 (approved):
LGTM!

PR #268, opened 2024-02-22
DOCSP 30174 clarifying estimatedCopiedByes and estimatedTotalBytes

ltran-mdb2:
## DESCRIPTION

clarifying that estimatedCopiedByes and estimatedTotalBytes are approximated before migration and should not be used as an indicator of progress. 

## STAGING

https://preview-mongodbltranmdb2.gatsbyjs.io/cluster-sync/DOCSP-30174/reference/api/progress/#successful-response

## JIRA

https://jira.mongodb.org/browse/DOCSP-30174

## BUILD LOG

https://workerpool-boxgs.mongodbstitch.com/pages/job.html?collName=queue&jobId=65d7783e7111165982411eb5

## SELF-REVIEW CHECKLIST

- [x] Does each file have 3-5 taxonomy facet tags?

### DIFF
--- a/source/includes/api/tables/progress-response.rst
+++ b/source/includes/api/tables/progress-response.rst
@@ -72,8 +72,8 @@
          ``mongosync`` approximates the estimated total number of bytes
          prior to migration and does not update this value during
          migration. This value does not reflect changes to the source
-         cluster during migration and should not be used as an indicator
-         of migration progress. 
+         cluster during migration and is not an accurate indicator of
+         migration progress. 
 
    * - ``collectionCopy``
        ``.estimatedCopiedBytes``
@@ -92,8 +92,8 @@
          ``mongosync`` approximates the estimated number of copied bytes
          prior to migration and does not update this value during
          migration. This value does not reflect changes to the source
-         cluster during migration and should not be used as an indicator
-         of migration progress. 
+         cluster during migration and is not an accurate indicator of
+         migration progress. 
 
 
    * - ``directionMapping``
--- a/source/includes/api/tables/progress-response.rst
+++ b/source/includes/api/tables/progress-response.rst
@@ -70,10 +70,10 @@
        .. note::
 
          ``mongosync`` approximates the estimated total number of bytes
-         prior to migration and does not update this value during
-         migration. This value does not reflect changes to the source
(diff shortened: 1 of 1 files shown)


jocelyn-mendez1 (approved):
Thank you @ltran-mdb2! This LGTM!

source/includes/api/tables/progress-response.rst
```diff
@@ -78,6 +87,15 @@
        and divide the result by the value of the ``estimatedTotalBytes`` field
        . Then, multiply the result by 100.
 
+       .. note::
+
+         ``mongosync`` approximates the estimated number of copied bytes
+         prior to migration and does not update this value during the
+         synchronization process. This value does not reflect changes
+         made to the source cluster during sync and is not an accurate
+         indicator of migration progress.
```
suspicious-mango, line 96: This should say something like:

mongosync only reflects the bytes copied by the current instance of mongosync. The value does not reflect bytes copied by other instances of mongosync and can become larger than estimatedTotalBytes due to retried operations. This value is not an accurate indicator of migration progress.

suspicious-mango (changes requested):
correction to estimatedCopiedBytes

source/includes/api/tables/progress-response.rst
suspicious-mango: Ahh apologies, I didn't see that the top part of the description for `estimatedCopiedBytes`. Then there is no need to restate "This value only reflects the bytes copied by the current instance of mongosync".

But the following section is incorrect and should be removed. mongosync does in fact compute the estimated number of copied bytes during the migration by adding up the size of documents that it copies. This value will reflect changes made to the source cluster during sync and can be higher than ``estimatedTotalBytes`` if documents are added to the source or lower if documents are deleted from the source.
```
         mongosync approximates the estimated number of copied bytes
         prior to migration and does not update this value during the
         synchronization process. This value does not reflect changes
         made to the source cluster during sync and is not an accurate
         indicator of migration progress.
```

ltran-mdb2: no worries, just made your suggested change!

suspicious-mango (approved):
LGTM.

PR #271, opened 2024-02-27
DOCSP-34589 clarify multiple mongosync limitations

ltran-mdb2:
## DESCRIPTION

correcting/adding to the mongosync multiple clusters limitations

## STAGING

https://preview-mongodbltranmdb2.gatsbyjs.io/cluster-sync/DOCSP-34589/reference/limitations/#multiple-clusters
https://preview-mongodbltranmdb2.gatsbyjs.io/cluster-sync/DOCSP-34589/reference/mongosync/#multiple-clusters

## JIRA

https://jira.mongodb.org/browse/DOCSP-34589

## BUILD LOG

https://workerpool-boxgs.mongodbstitch.com/pages/job.html?collName=queue&jobId=65de1fb17111165982c9ea2a

## SELF-REVIEW CHECKLIST

- [x] Does each file have 3-5 taxonomy facet tags?

### DIFF
--- a/source/reference/limitations.txt
+++ b/source/reference/limitations.txt
@@ -137,6 +137,8 @@ Multiple Clusters
 - One cluster cannot simultaneously be a source cluster in one
   ``mongosync`` instance and a destination cluster in another
   ``mongosync`` instance. 
+- To sync a source cluster to multiple destination clusters, use
+   one ``mongosync`` instance for each destination cluster. 
 
 .. _c2c-filtering-limitations:
 
--- a/source/reference/limitations.txt
+++ b/source/reference/limitations.txt
@@ -129,6 +129,8 @@ Reversing
   Ensure that unique indexes exist on all shards before reversing.
 - .. include:: /includes/fact-reverse-limitation.rst
 
+.. _multiple-clusters-limitations:
+
 Multiple Clusters
 -----------------
 
@@ -137,8 +139,6 @@ Multiple Clusters
 - One cluster cannot simultaneously be a source cluster in one
   ``mongosync`` instance and a destination cluster in another
   ``mongosync`` instance. 
--  To sync a source cluster to multiple destination clusters, use one
-   ``mongosync`` instance for each destination cluster. 
 
 .. _c2c-filtering-limitations:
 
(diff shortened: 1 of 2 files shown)


source/reference/limitations.txt
```diff
@@ -132,10 +132,13 @@ Reversing
 Multiple Clusters
 -----------------
 
-- Syncing multiple source clusters to one destination cluster isn't
-  supported.
-- Syncing one source cluster to many destination clusters isn't
-  supported.
+- ``mongosync`` does not support syncing multiple source clusters to one
+  destination cluster. 
+- One ``mongosync`` instance cannot sync a source cluster to many
```
mdb-ashley, line 137: Overall this LGTM, however -- do we want to specify here that you *can* run multiple mongosyncs from the same source cluster?

ltran-mdb2: I think it makes sense to specify! I added a note under the limitation

source/reference/limitations.txt
```diff
@@ -132,10 +132,19 @@ Reversing
...
```
rohshar, line 137: I don't think that this point is needed. Its a pretty obvious limitation that I don't think needs to be called out. I think it does make sense to call out the thing in the note below somewhere.

ltran-mdb2: Because this page previously said that mongosync **cannot** support 1 to N replication so I think it might be important to note this, at least for a while, in case users become confused with the change

rohshar: To me, it doesn't seem that users would be attached to that previous limitation, or remember it. I think its more confusing to add that point, and I'd rather the docs just be as clear as possible in their current form.

ltran-mdb2: That makes sense, I removed the bullet. Could you give this another review when you get a chance?

rohshar (changes requested):
I think we should document somewhere that its possible to sync from one source cluster to multiple destination clusters

ltran-mdb2:
Added the note back as a bullet! What do you think @rohshar ?

rohshar:
Its a little confusing because the new bullet is listed in the limitations page, but its not a limitation. Is there a better place to put it?

ltran-mdb2:
@rohshar I added the information on the mongosync page and linked to the multiple clusters limitation section. What do you think?

rohshar (approved):

mdb-ashley (approved):
LGTM! Thanks, Lauren.

PR #287, opened 2024-03-18
DOCSP-23937 Fix Role Issue on Mongosync Start Page

kanchana-mongodb:
- [JIRA](https://jira.mongodb.org/browse/DOCSP-23937): Document for [setUserWriteBlockMode](https://www.mongodb.com/docs/master/reference/privilege-actions/#mongodb-authaction-setUserWriteBlockMode) and [bypassWriteBlockingMode](https://www.mongodb.com/docs/master/reference/privilege-actions/#mongodb-authaction-bypassWriteBlockingMode) specify that these roles should be applied to the cluster resource, but current example in [doc](https://www.mongodb.com/docs/cluster-to-cluster-sync/current/reference/api/start/#user-write-blocking) shows these applied on the db/collection. Fixes incorrect documentation.
- [Stage](https://preview-mongodbkanchanamongodb.gatsbyjs.io/cluster-sync/DOCSP-23937/reference/api/start/#user-write-blocking/) 
- [Build](https://workerpool-boxgs.mongodbstitch.com/pages/job.html?collName=queue&jobId=65f8afba6fc5f04850b6162a)

### DIFF
--- a/source/reference/api/start.txt
+++ b/source/reference/api/start.txt
@@ -41,7 +41,7 @@ To set a custom role for the ``mongosync`` user:
       db.adminCommand( {
          createRole: "reverseSync",
          privileges: [ {
-             resource: { db: "", collection: "" },
+             resource: { cluster: "true" },
              actions: [ "setUserWriteBlockMode", "bypassWriteBlockingMode" ]
          } ],
          roles: []
--- a/source/reference/api/start.txt
+++ b/source/reference/api/start.txt
@@ -27,36 +27,18 @@ State
 To use the ``start`` endpoint, ``mongosync`` must be in the ``IDLE``
 state.
 
-User Write Blocking
-~~~~~~~~~~~~~~~~~~~
+Permissions
+~~~~~~~~~~~
 
-.. include:: /includes/fact-write-blocking-requirement.rst
+The user specified in the ``mongosync`` connection string must have the
+required permissions on the source and destination clusters. Refer to
+:ref:`c2c-permissions-and-roles` to ensure that the user has the correct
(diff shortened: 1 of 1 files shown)


kennethdyer (approved):
Thanks @kanchana-mongodb, This LGTM!

rohshar (commented):
I think this becomes overly complicated in trying to tell the users how to create custom roles instead of just linking to the reserve part of the user permissions section: https://www.mongodb.com/docs/cluster-to-cluster-sync/current/reference/permissions/

I'm not sure why this was added initially, and since it isn't totally correct, I'm not sure if anyone has used it.

@tfogo any thoughts?

tfogo:
@rohshar For a quick start I think it is important to have all the steps laid out in a way that users can easily copy and paste. So I think it is good practice to have a command. Linking to the permissions doc works, but having the command is better.

rohshar:
My issue is that the commands listed in the quickstart are for creating a custom role, but seem to have been wrong before. I think rather than prompting users to create custom roles, it is probably better for them to use the built in roles, hence linking to the permissions doc.

rohshar (changes requested):
Synced with Tim. Since this isn't in the quickstart, we shouldn't add documentation that is totally different than what is present in https://www.mongodb.com/docs/cluster-to-cluster-sync/current/reference/permissions/

Instead, I think we should only mention that they can check the permissions needed in the permissions page and create a custom role with those here.

tfogo:
Yeah apologies, I saw the file name was "start" and thought it was the quick start.

rohshar (approved):

PR #294, opened 2024-04-03
DOCSP-37693-destination-sharded-cluster-with-arbiter-not-supported-mo…

mmaville-mdb:
## DESCRIPTION
Adding limitation that mongosync doesn't support syncing into a sharded cluster topology with one or more arbiters.

## JIRA
https://jira.mongodb.org/browse/DOCSP-37693

## BUILD LOG
https://workerpool-boxgs.mongodbstitch.com/pages/job.html?collName=queue&jobId=660daa243a5920a291bcf57a

### DIFF
--- a/source/reference/limitations.txt
+++ b/source/reference/limitations.txt
@@ -76,6 +76,8 @@ Sharded Clusters
 
 - ``mongosync`` doesn't support sync from a sharded cluster
   to a replica set.
+- ``mongosync`` doesn't support syncing into a sharded cluster 
+  topology with one or more arbiters.
 - Sync from a replica set to a sharded cluster has the following
   limitations:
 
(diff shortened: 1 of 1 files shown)


mmaville-mdb:
> What does "topology" gain us in terms of clarity here? Can we just drop topology and mean the same thing?

@sarah-olson-mongodb I thought about that for a bit. Maybe this will help inform: I'm going off of the [Production Considerations](https://www.mongodb.com/docs/manual/core/transactions-sharded-clusters/#multiple-shards) for multiple shards, where it states "On a sharded cluster, transactions that span multiple shards will error and abort if any involved shard contains an arbiter."

I wasn't sure if sharded cluster was specific enough, so I hoped leaving in topology would invoke the consideration all participating members. Perhaps it's redundant though.

sarah-olson-mongodb:
That makes sense, @mmaville-mdb! Thanks for thinking it through!

tfogo (approved):
LGTM

PR #295, opened 2024-04-03
DOCSP-34591 Add errorDescription and success keys to progress response

kanchana-mongodb:
- [JIRA](https://jira.mongodb.org/browse/DOCSP-34591)
- [Stage](https://preview-mongodbkanchanamongodb.gatsbyjs.io/cluster-sync/DOCSP-34591/reference/api/progress/#error-response)
- [Build](https://workerpool-boxgs.mongodbstitch.com/pages/job.html?collName=queue&jobId=660dbe2a3a5920a291c9aa74)

### DIFF
--- a/source/includes/api/tables/progress-response.rst
+++ b/source/includes/api/tables/progress-response.rst
@@ -114,11 +114,11 @@
 
    * - ``mongosyncID``
      - string
-     - The identifier string for the ``mongosync`` instance.
+     - Identifier string for the ``mongosync`` instance.
 
    * - ``coordinatorID``
      - string
-     - The identifier string for the coordinator instance.
+     - Identifier string for the coordinator instance.
 
        - When ``mongosync`` is coordinated by another instance, this field shows
          the identifier string for the coordinator instance.
@@ -128,4 +128,3 @@
 
        - When ``mongosync`` starts up, this field returns ``null`` until
          ``mongosync`` identifies the coordinator.
-
--- a/source/reference/api/progress.txt
+++ b/source/reference/api/progress.txt
@@ -37,8 +37,7 @@ The response object contains 2 top-level fields, ``success`` and
 ``progress``. 
 
 The ``success`` field contains the status of the ``progress`` command. 
-Value is ``true`` if the command succeeds and ``false`` if the command
-fails. 
+Value is ``true`` if always ``true``. 
 
 If ``mongosync`` successfully gets the status of the sync process, 
 all response fields are wrapped in the top-level ``progress`` object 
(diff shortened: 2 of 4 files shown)


mmcclimon (changes requested):
One tweak to make, plus two other changes on the page (https://preview-mongodbkanchanamongodb.gatsbyjs.io/cluster-sync/DOCSP-34591/reference/api/progress/#response-1) that aren't in this .rst file:

First, just above the error response table:
> If mongosync encounters an error, the progress endpoint returns the following field:

This should be "fields" (plural).

Second: the `success` boolean field is also returned for non-error responses. I don't know if we want to document this in the table, but I think we should at least update the sample successful response to this (note the top-level "success" key):

```
{
   "success":  true,
   "progress":
      {
         "state":"RUNNING",
         "canCommit":true,
         "canWrite":false,
         "info":"change event application",
         "lagTimeSeconds":0,
         "collectionCopy":
            {
               "estimatedTotalBytes":694,
               "estimatedCopiedBytes":694
            },
         "directionMapping":
            {
               "Source":"cluster0: localhost:27017",
               "Destination":"cluster1: localhost:27018"
            }
      }
}
```

Thanks!

source/includes/api/tables/progress-response.rst
```diff
@@ -129,3 +129,7 @@
        - When ``mongosync`` starts up, this field returns ``null`` until
          ``mongosync`` identifies the coordinator.
 
+   * - ``success``
+     - boolean
+     - Status of the ``progress`` command. Value is ``true`` if the 
+       command succeeds and ``false`` if the command fails.
```
mmcclimon, line 135: This table item is a little misleading. This table is introduced by the sentence
> If mongosync successfully gets the status of the sync process, all response fields are wrapped in a top-level `progress` object with the following fields:

The `success` key is a top-level field, not part of the `progress` object. That is: a successful response looks like this:
```javascript
{
  success: true,
  progress: {
    // All the other fields as in the table
  }
}
```

So, this description is fine, but it's a top-level field and not part of the `progress` object.

mmcclimon (changes requested):
One more tweak; thanks!

sarah-olson-mongodb (approved):
Thanks @kanchana-mongodb! LGTM

source/reference/api/progress.txt
```diff
@@ -33,8 +33,15 @@ The ``progress`` endpoint returns either an updated status or an error.
 Successful Response
 ~~~~~~~~~~~~~~~~~~~
 
+The response object contains 2 top-level fields, ``success`` and
+``progress``. 
+
+The ``success`` field contains the status of the ``progress`` command. 
+Value is ``true`` if the command succeeds and ``false`` if the command
+fails. 
+
```
mmcclimon, line 42: (nit) Since this is in the "successful response" section, we could maybe simplify by saying here that `success` will always be true here.

mmcclimon (approved):
LGTM % 1 small suggestion; that isn't really a technical point but a clarity one, so happy to approve without needing another review. Thanks!

PR #329, opened 2024-06-13
DOCSP-40106 C2C Beta Disclaimer

ajhuh-mdb:
## DESCRIPTION 
Adds a notice for users that they must accept the disclaimer when they first run the `mongosync` beta binary. 

## STAGING 
https://preview-mongodbajhuhmdb.gatsbyjs.io/cluster-sync/DOCSP-40106-beta-disclaimer/reference/beta-program/#get-started

## JIRA 
https://jira.mongodb.org/browse/DOCSP-40106

## BUILD
https://workerpool-boxgs.mongodbstitch.com/pages/job.html?collName=queue&jobId=666b52b1ebce19f314aa11d2

### DIFF
--- a/source/reference/beta-program.txt
+++ b/source/reference/beta-program.txt
@@ -23,6 +23,9 @@ Get Started
 To request access to the {+c2c-beta-program-short+} binaries, please file a 
 `HELP ticket <https://jira.mongodb.org/browse/HELP>`_.
 
+When you first run the {+c2c-beta-program-short+} binary, you must accept the 
+following disclaimer:
+
 .. warning:: Disclaimer
 
    MongoDB restricts access to beta features via the beta build. Your use of the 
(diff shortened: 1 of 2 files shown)


source/reference/beta-program.txt
```diff
@@ -23,25 +23,36 @@ Get Started
...
+following disclaimer:
+
+.. code-block:: shell
+   :copyable: false
+  
+   MongoDB restricts access to beta features via the beta build. Your use 
+   of the beta is governed by the language specified in the Cloud 
+   Subscription Agreement, Cloud Terms of Service, or other applicable 
+   agreement between you and MongoDB with respect to your use of 
+   ``mongosync``.
```
kennethdyer, line 36: Question: Does the shell output include the double ticks?  I imagine it shouldn't but we should follow whatever the user sees in the output.

kennethdyer (approved):
Thanks for this @ajhuh-mdb, LGTM with a question.

source/reference/beta-program.txt
ajhuh-mdb: Ah that was left over from when this used to be an admonition instead of a code block. Updating.

ajhuh-mdb:
LGTM'd by Revathy via Slack.

PR #340, opened 2024-07-08
DOCSP-40742-disaster-recovery-page-updates

ajhuh-mdb:
## DESCRIPTION 
- Wording updates to disaster recovery page + updated table.

## STAGING 
https://preview-mongodbajhuhmdb.gatsbyjs.io/cluster-sync/DOCSP-40742-disaster-recovery/reference/disaster-recovery/

## JIRA 
https://jira.mongodb.org/browse/DOCSP-40742

## BUILD
https://workerpool-boxgs.mongodbstitch.com/pages/job.html?collName=queue&jobId=668c54b9fb8bd13b8438e58c

PR #342, opened 2024-07-08
DOCSP-40742-disaster-recovery-page-updates (#340)

ajhuh-mdb:
Ports #340 to v1.7 

BUILD: https://workerpool-boxgs.mongodbstitch.com/pages/job.html?collName=queue&jobId=668c58d4fb8bd13b8439bfc7

### DIFF
--- a/source/reference/disaster-recovery.txt
+++ b/source/reference/disaster-recovery.txt
@@ -60,6 +60,12 @@ the following changes that mimic the sync finalization process of the
 
        If synchronization was started with write blocking, disable the Write
        Block Mode.
+   * - Dummy Indexes 
+     - In some cases, synchronization may create dummy indexes on the 
+       destination to support writes on sharded or collated collections. 
+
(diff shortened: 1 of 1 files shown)


source/reference/disaster-recovery.txt
```diff
@@ -141,3 +153,22 @@ To enable writes, update :dbcommand:`setUserWriteBlockMode`:
        }
    )
 
+.. _c2c-dr-capped-collections:
+
+Capped Collections
```
jeff-allen-mongo, line 158: Hey @ajhuh-mdb I don't see this change as part of the original [PR](https://github.com/mongodb/docs-cluster-to-cluster-sync/pull/340/files). I'm guessing this was added to resolve a build error, but just want to confirm before I merge that it wasn't added by mistake. Thanks!

ajhuh-mdb: Yep, that's correct! There was a build error that required this section to be added. 👍

PR #375, opened 2024-08-06
C2C Beta Docs  (#370)

ajhuh-mdb:
Merging c2c beta feature branch

STAGING: https://preview-mongodbajhuhmdb.gatsbyjs.io/cluster-sync/master/reference/beta-program/

BUILD: https://workerpool-boxgs.mongodbstitch.com/pages/job.html?collName=queue&jobId=66b2632d824f175afd281c56

### DIFF
--- a/source/includes/fact-valid-namespace-remap.rst
+++ b/source/includes/fact-valid-namespace-remap.rst
@@ -1,0 +1,33 @@
+
+The following restrictions apply to namespace remapping:
+
+- Namespace remapping doesn't permit writing to the ``system``, ``config``, ``admin``,
+  or ``local`` databases, or writing to internal databases used by ``mongosync``.
+
+- The database name on the destination cluster must be valid under Windows
+  restrictions.
+
+  For more information, see :limit:`Restrictions on Database Names for Windows`.
+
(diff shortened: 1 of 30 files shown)


source/includes/fact-valid-namespace-remap.rst
```diff
@@ -0,0 +1,33 @@
+
+The following restrictions apply to namespace remapping:
```
jtcovan, line 2: Is it possible to add this limitation:

```
You can't specify a namespace remap and set the ``reversible`` flag to 
  ``true``.
```

If not, no worries. We can do this in a follow up ticket after the release.

jtcovan (commented):
LGTM, thanks so much! I had one comment about adding a limitation for namespace remapping but totally okay if it's better for me to put that in a follow up ticket.

source/includes/fact-valid-namespace-remap.rst
ajhuh-mdb: Just to make sure we're not adding too much content that isn't reviewed by the rest of the writing team, could you put this in a fast-follow ticket?

PR #377, opened 2024-08-07
DOCSP-41999-mongosync-disaster-recovery

ajhuh-mdb:
## DESCRIPTION
Changes discussed with Maria van Keulen and Rohan S.: 
- Taking a stronger stance against using `mongosync` for disaster recovery use cases, warning of stale/inconsistent data + non-atomic transactions
- Removing [configuration change recommendations
](https://www.mongodb.com/docs/cluster-to-cluster-sync/current/reference/disaster-recovery/#write-blocking:~:text=To%20prepare%20the%20destination%20cluster%20to%20replace%20the%20source%20cluster%2C%20make%20the%20following%20changes%20that%20mimic%20the%20sync%20finalization%20process%20of%20the%20commit%20endpoint) 

Since it seemed odd to to have a standalone page on considerations before using `mongosync` I opted to add the warning about DR in the [mongosync behavior page's consistency section](https://www.mongodb.com/docs/cluster-to-cluster-sync/current/reference/mongosync/mongosync-behavior/#destination-clusters). This section already seems to include the requested information on transactions & eventual consistency. Here I:
- Add a warning that we don't recommend using `mongosync` for disaster recovery under the note on eventual consistency
- Delete the disaster recovery page & redirect users to the mongosync behavior page section
- Create "Considerations for Continuous Sync" section in about mongosync page

## STAGING 
- [mongosync Behavior (consistency)](https://preview-mongodbajhuhmdb.gatsbyjs.io/cluster-sync/DOCSP-41999-mongosync-dr/reference/mongosync/mongosync-behavior/#consistency)
- [faq (write blocking link swap)](https://preview-mongodbajhuhmdb.gatsbyjs.io/cluster-sync/DOCSP-41999-mongosync-dr/faq/#can-i-perform-reads-or-writes-to-my-destination-cluster-while-mongosync-is-syncing-)
- [Considerations for Continuous Sync](https://preview-mongodbajhuhmdb.gatsbyjs.io/cluster-sync/DOCSP-41999-mongosync-dr/reference/mongosync/mongosync-behavior/#considerations-for-continuous-sync)
## JIRA
https://jira.mongodb.org/browse/DOCSP-41999


ajhuh-mdb:
Here's the `mut-redirect` outputs for all the disaster recovery page redirects: 

> Redirect 301 /docs/cluster-to-cluster-sync/v1.7/reference/disaster-recovery https://www.mongodb.com/docs/cluster-to-cluster-sync/v1.7/reference/mongosync/mongosync-behavior/#consistency
Redirect 301 /docs/cluster-to-cluster-sync/1.8/reference/disaster-recovery https://www.mongodb.com/docs/cluster-to-cluster-sync/1.8/reference/mongosync/mongosync-behavior/#consistency
Redirect 301 /docs/cluster-to-cluster-sync/master/reference/disaster-recovery https://www.mongodb.com/docs/cluster-to-cluster-sync/master/reference/mongosync/mongosync-behavior/#consistency

jocelyn-mendez1 (approved):
Thanks @ajhuh-mdb, LGTM!

source/reference/mongosync/mongosync-behavior.txt
```diff
@@ -140,11 +140,17 @@ Consistency
...
 
 ``mongosync`` supports :term:`eventual consistency` on the destination
-cluster. Read Consistency is not guaranteed on the destination cluster,
-the source and destination clusters may differ at a given point in
+cluster. Read consistency is not guaranteed on the destination cluster.
+The source and destination clusters may differ at a given point in
 time. However, if new writes are halted on the source cluster, the
 source and destination clusters will eventually hold the same data.
 
+.. warning::
```
mvankeulen94, line 148: BTW here's the blurb we had been discussing to share with a customer who wanted to use mongosync for continuous sync:

> mongosync must [commit](https://www.mongodb.com/docs/cluster-to-cluster-sync/current/reference/api/commit/) for the destination cluster to have an up to date, consistent view of the source cluster’s data. If a disaster occurs and shuts down the source cluster before mongosync can commit, the destination cluster will not have a consistent snapshot of the source data. For any continuous synchronization use cases with mongosync, ensure mongosync commits before cutting over from the source to the destination.
When used with sharded clusters, balancers need to be paused on both the source and the destination for the lifetime of the sync. If mongosync is used for continuous synchronization, this means that the balancers will remain disabled on either cluster.

Can we have a 'considerations for continuous sync' section?

mvankeulen94 (changes requested):
Will give Tim a heads up on this change so he can weigh in. Thanks for working on this!

source/reference/mongosync/mongosync-behavior.txt
ajhuh-mdb: I'll slot this in! Thanks for sending the copy over.

mvankeulen94: Thanks! If we could also add ", and there may be torn transactions." to the "[...] will not have a consistent snapshot of the source data" sentence that would be great

mvankeulen94:
@ajhuh-mdb BTW quick update, we should hold off on pushing this until we have circulated this messaging internally to our account teams to make sure there are no issues with this public messaging.

source/about-mongosync.txt
```diff
@@ -157,6 +157,30 @@ If the reverse sync is successful, ``mongosync`` enters the ``RUNNING`` state.
 The synchronization continues in the reverse direction from the original 
 sync job. 
 
+.. _mongosync-considerations:
+
+Considerations for Continuous Sync
+----------------------------------
+
+Avoid using ``mongosync`` for disaster recovery for the following reasons:
```
mvankeulen94, line 165: Instead of having an explicit call to action, we can instead have something like this as the first sentence:

"mongosync can be run periodically against source and destination clusters to continuously synchronize between the two. Below are considerations to keep in mind if using mongosync in this way:"

source/about-mongosync.txt
```diff
@@ -157,6 +157,30 @@ If the reverse sync is successful, ``mongosync`` enters the ``RUNNING`` state.
...
+
+- For the destination cluster to have an up to date, consistent view of the 
+  source cluster’s data, ``mongosync`` must :ref:`commit <c2c-api-commit>` . 
+  If a disaster occurs and shuts down the source cluster before ``mongosync``
```
mvankeulen94, line 169: We can update to:
"If the source cluster shuts down before mongosync can commit (e.g., due to a disaster scenario), the cluster won't [...]"

mvankeulen94 (changes requested):
still soliciting other approvals on text

source/about-mongosync.txt
ajhuh-mdb: updated!

source/about-mongosync.txt
```diff
@@ -162,13 +162,15 @@ sync job.
...
+following considerations:
 
 - For the destination cluster to have an up to date, consistent view of the 
   source cluster’s data, ``mongosync`` must :ref:`commit <c2c-api-commit>` . 
-  If a disaster occurs and shuts down the source cluster before ``mongosync`` 
-  commits, the destination cluster won't have a consistent snapshot of the 
-  source data and transactions may be torn. 
+  If the source cluster shuts down before ``mongosync`` can commit, such as in 
+  a disaster scenario, the destination cluster won't have a consistent snapshot 
+  of the source data and transactions may be torn.
```
mvankeulen94, line 173: to "may be torn", we can add ", i.e., not be applied atomically", in case users haven't heard 'torn transactions' before

source/about-mongosync.txt
```diff
@@ -162,13 +162,15 @@ sync job.
 Considerations for Continuous Sync
 ----------------------------------
 
-Avoid using ``mongosync`` for disaster recovery for the following reasons:
+You can periodically run ``mongosync`` against source and destination clusters 
+in continuous synchronization. Using ``mongosync`` for continuous sync has the
```
mvankeulen94, line 166: wordsmithing nit: instead of the "Using" sentence, we can say "Please bear the following considerations in mind when using mongosync for continuous synchronization:"

mvankeulen94 (changes requested):
just a couple minor wordsmithing changes. CC @rohshar could you also please take a look? Once we're happy with the draft, I will circulate the draft internally to make sure the messaging looks good. we should hold off on pushing until we've gotten the go-ahead from internal stakeholders

source/about-mongosync.txt
```diff
@@ -157,6 +157,32 @@ If the reverse sync is successful, ``mongosync`` enters the ``RUNNING`` state.
...
+----------------------------------
+
+You can periodically run ``mongosync`` against source and destination clusters 
+in continuous synchronization. Using ``mongosync`` for continuous sync has the 
+following considerations:
+
+- For the destination cluster to have an up to date, consistent view of the 
+  source cluster’s data, ``mongosync`` must :ref:`commit <c2c-api-commit>` . 
+  If the source cluster shuts down before ``mongosync`` can commit, such as in 
+  a disaster scenario, the destination cluster won't have a consistent snapshot
```
rohshar, line 172: may not have. Its not guaranteed that it won't have

source/about-mongosync.txt
```diff
@@ -157,6 +157,32 @@ If the reverse sync is successful, ``mongosync`` enters the ``RUNNING`` state.
 The synchronization continues in the reverse direction from the original 
 sync job. 
 
+.. _mongosync-considerations:
+
+Considerations for Continuous Sync
+----------------------------------
+
+You can periodically run ``mongosync`` against source and destination clusters
```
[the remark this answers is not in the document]
mvankeulen94: Thanks @rohshar . What are the steps that a user needs to take in order to continuously run mongosync?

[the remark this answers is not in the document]
rohshar: They just need to have mongosync run without interruption. 

In practice, this likely means that they run mongosync and then monitor/alert if it goes down. If it goes down then they can start mongosync again and depending on the error it'll be able to catch back up. There are some errors that mongosync can't recover from.

source/about-mongosync.txt
ajhuh-mdb: It's against our documentation style to use abbreviations like `i.e.` since it's hard for international audiences to parse. Will opt for similar verbiage as in the [mongosync behavior page](https://www.mongodb.com/docs/cluster-to-cluster-sync/current/reference/mongosync/mongosync-behavior/#consistency:~:text=Transactions%20may%20appear%20non%2Datomically%20on%20the%20destination%20cluster):

> Transactions may appear non-atomically on the destination cluster

source/about-mongosync.txt
ajhuh-mdb: Will adjust this to align with our docs' style and tone.

source/about-mongosync.txt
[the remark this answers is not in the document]
ajhuh-mdb: Thanks for the explanation. For now, I'll remove that first sentence since it seems it's technically inaccurate and just introduce the section with the "Consider the following..." sentence.

source/about-mongosync.txt
```diff
@@ -157,6 +157,31 @@ If the reverse sync is successful, ``mongosync`` enters the ``RUNNING`` state.
 The synchronization continues in the reverse direction from the original 
 sync job. 
 
+.. _mongosync-considerations:
+
+Considerations
```
rohshar, line 162: Why did we change this from considerations about continuous sync?

source/faq.txt
```diff
@@ -30,7 +30,7 @@ destination cluster as long as the source cluster doesn’t include a namespace
 with the same name. If you write to a synced namespace before issuing a
```
rohshar, line 30: Above where it says:

> You can perform reads during synchronization if ``canCommit`` is ``true``. 
However, the data that you read is 
:term:`eventually consistent <eventual consistency>`, meaning that you might not 
always read the latest writes.

Some clarifications: you can read at any time, but you may not get the latest writes due to eventual consistency. If you wait until committed, you will get the up to date writes.

rohshar (changes requested):
Generally good, a couple smaller things.

source/about-mongosync.txt
ajhuh-mdb: hm looks like that's a mistake on my end — fixing.

source/faq.txt
ajhuh-mdb: updated

rohshar:
Happy to merge once Maria circulates it and gives you the go ahead.

source/reference/mongosync/mongosync-behavior.txt
```diff
@@ -140,16 +140,21 @@ Consistency
...
-time. However, if new writes are halted on the source cluster, the
-source and destination clusters will eventually hold the same data.
+cluster. Read consistency is not guaranteed on the destination cluster until 
+commit. Before committing, the source and destination clusters may differ at a 
+given point in time. However, if new writes are halted on the source cluster, 
+the source and destination clusters will eventually hold the same data.
+
+.. warning:: 
+   
+     We do not recommend using ``mongosync`` for disaster recovery. To learn
```
mvankeulen94, line 150: I would recommend removing this. Customers will decide for themselves whether running mongosync continuously makes sense for their purposes, given the risks. Also, right now customers are encountering this page because they are searching for "disaster recovery mongodb", when the go to solution should actually be replication with automated failover in many cases. I'd like to exclude the string "disaster recovery" from this page.

mvankeulen94 (changes requested):

source/reference/mongosync/mongosync-behavior.txt
ajhuh-mdb: Gotcha. Updating.

ajhuh-mdb: In that case, should we be explicitly mentioning that they should use replication w/ automated failover in the case of a disaster scenario? With us deleting the disaster recovery page, we never really provide a solution for it.

rohshar: We don't really have a solution for it

mvankeulen94:
Awesome, thank you! Will circulate now and keep you posted if there are any further comments.

source/about-mongosync.txt
```diff
@@ -157,6 +157,31 @@ If the reverse sync is successful, ``mongosync`` enters the ``RUNNING`` state.
...
+  
+  - For any continuous synchronization use cases with ``mongosync``, ensure that 
+    ``mongosync`` commits before cutting over from the source to the 
+    destination.
+
+- When used with sharded clusters, you must pause :ref:`balancers 
+  <sharding-balancing>` on both the source and the destination for the lifetime 
+  of the sync.
+  
+  - If you use ``mongosync`` for continuous synchronization, the balancers will
```
mvankeulen94, line 182: thinking about this some more, I don't think we need this sub-bullet. I got the feedback that the "will" wording here is confusing.

we could instead have a follow-on sentence to the bullet "When used with sharded clusters" along these lines:

"Running mongosync continuously means that this condition will need to remain true indefinitely." CC @rohshar lmk your thoughts

source/about-mongosync.txt
```diff
@@ -157,6 +157,31 @@ If the reverse sync is successful, ``mongosync`` enters the ``RUNNING`` state.
...
+  If the source cluster shuts down before ``mongosync`` can commit, such as in 
+  a disaster scenario, the destination cluster might not have a consistent 
+  snapshot of the source data, and transactions might not appear atomically on 
+  the destination cluster.
+  
+  - For any continuous synchronization use cases with ``mongosync``, ensure that 
+    ``mongosync`` commits before cutting over from the source to the 
+    destination.
+
+- When used with sharded clusters, you must pause :ref:`balancers
```
mvankeulen94, line 178: feedback from Ritesh: "pause" -> "stop"

source/about-mongosync.txt
```diff
@@ -157,6 +157,31 @@ If the reverse sync is successful, ``mongosync`` enters the ``RUNNING`` state.
...
+  snapshot of the source data, and transactions might not appear atomically on 
+  the destination cluster.
+  
+  - For any continuous synchronization use cases with ``mongosync``, ensure that 
+    ``mongosync`` commits before cutting over from the source to the 
+    destination.
+
+- When used with sharded clusters, you must pause :ref:`balancers 
+  <sharding-balancing>` on both the source and the destination for the lifetime 
+  of the sync.
```
mvankeulen94, line 180: feedback from Ritesh: add "until commit" to "lifetime of the sync"

mvankeulen94: CC @rohshar

rohshar: fine with me

source/about-mongosync.txt
ajhuh-mdb: On second read, it might be redundant to state this again, so I'm leaning towards just taking out the sub-bullet and keeping it with just the first sentence. With just the top bullet, as is, it does emphasize that the balancers need to be stopped on both the source & destination clusters.

source/faq.txt
```diff
@@ -20,20 +20,20 @@ Support.
 Can I perform reads or writes to my destination cluster while ``mongosync`` is syncing?
 ---------------------------------------------------------------------------------------
 
-You can perform reads during synchronization if ``canCommit`` is ``true``. 
-However, the data that you read is 
-:term:`eventually consistent <eventual consistency>`, meaning that you might not 
-always read the latest writes.
+You can perform reads during synchronization at any time. However, the data 
+that you read is :term:`eventually consistent <eventual consistency>`, meaning 
+that you might not always read the latest writes. For up-to-date reads, wait
```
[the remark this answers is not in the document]
mvankeulen94: We should also remove any implication that reading from the destination prior to commit will just result in a stale read, given the above

source/reference/mongosync/mongosync-behavior.txt
```diff
@@ -140,16 +140,16 @@ Consistency
 ~~~~~~~~~~~
 
 ``mongosync`` supports :term:`eventual consistency` on the destination
-cluster. Read Consistency is not guaranteed on the destination cluster,
-the source and destination clusters may differ at a given point in
-time. However, if new writes are halted on the source cluster, the
-source and destination clusters will eventually hold the same data.
+cluster. Read consistency is not guaranteed on the destination cluster until 
+commit. Before committing, the source and destination clusters may differ at a 
+given point in time. However, if new writes are halted on the source cluster,
```
mvankeulen94, line 145: I think we should add "upon commit" to "eventually hold the same data" CC @rohshar

source/reference/mongosync/mongosync-behavior.txt
```diff
@@ -140,16 +140,16 @@ Consistency
...
-cluster. Read Consistency is not guaranteed on the destination cluster,
-the source and destination clusters may differ at a given point in
-time. However, if new writes are halted on the source cluster, the
-source and destination clusters will eventually hold the same data.
+cluster. Read consistency is not guaranteed on the destination cluster until 
+commit. Before committing, the source and destination clusters may differ at a 
+given point in time. However, if new writes are halted on the source cluster, 
+the source and destination clusters will eventually hold the same data.
 
 Individual writes may be reordered or combined as they are relayed from
```
mvankeulen94, line 148: For this section starting with "Individual writes" and the below section starting with "Transactions might not", these are specifically for pre-commit mongosync, right? Can we label this more clearly? CC @rohshar

mvankeulen94 (changes requested):
have been thinking about this change some more in response to feedback & threads that occurred in the meantime, and I think there are some additional clarifications that we can make

source/faq.txt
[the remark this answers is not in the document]
rohshar: I think that context would be helpful as well.

source/reference/mongosync/mongosync-behavior.txt
rohshar: After commit, mongosync won't be doing any other writes, so just the client should be doing the writes.

source/reference/mongosync/mongosync-behavior.txt
rohshar: That is fine with me. I think the idea is that the documents should eventually match if no new writes are happening on the source, even before commit, but other configuration may not. It is reasonable not to want to publicize this to clients.

mvankeulen94 (changes requested):

source/faq.txt
```diff
@@ -20,20 +20,20 @@ Support.
...
-You can perform reads during synchronization if ``canCommit`` is ``true``. 
-However, the data that you read is 
-:term:`eventually consistent <eventual consistency>`, meaning that you might not 
-always read the latest writes.
+You can perform reads during synchronization at any time. However, the data 
+that you read is :term:`eventually consistent <eventual consistency>`, meaning 
+that you might not always read the latest writes. For up-to-date reads, wait 
+for the migration to commit.
 
 During an ongoing sync, you can write to any non-synced namespaces in the
```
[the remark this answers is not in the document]
rohshar: If you're using namespace filtering some namespaces may not be synced and can be considered non-synced namespaces

source/faq.txt
```diff
@@ -20,20 +20,20 @@ Support.
...
 with the same name. If you write to a synced namespace before issuing a 
 :ref:`commit <c2c-api-commit>` and while ``canWrite`` is ``false``, the behavior 
 is undefined. To avoid this undefined behavior, you can enable 
-:ref:`write blocking <c2c-dr-write-blocking>`.
+:ref:`write blocking <c2c-write-blocking>`.
 
-To check the value of ``canCommit`` or ``canWrite``, call the :ref:`progress 
-<c2c-api-progress>` API endpoint.
+To check the value of ``canWrite``, call the :ref:`progress <c2c-api-progress>` 
+API endpoint.
```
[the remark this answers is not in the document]
rohshar: That change sounds reasonable to me

source/reference/mongosync/mongosync-behavior.txt
ajhuh-mdb: updated to just remove the sentence.

source/reference/mongosync/mongosync-behavior.txt
ajhuh-mdb: updated.

source/faq.txt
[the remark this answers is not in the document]
ajhuh-mdb: updated

source/faq.txt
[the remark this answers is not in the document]
mvankeulen94: Gotcha, thanks. I'm not following what it means to have a non-synced namespace in the destination cluster - does that mean the customer can issue random writes to the destination on namespaces not included in the mongosync, as long as it happens after mongosync begins?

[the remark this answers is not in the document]
rohshar: There are two cases:
- A GA sync. If there is any data on the destination at the start of the sync, the migration will error. If clients write to a non-included namespace while the migration is going on, this should not cause problems with mongosync. They'd have to be really careful to not write to a synced namespace or they will cause data inconsistency. It does make it harder to verify, for example, when 1.9 goes out if they want to use the embedded verifier, if they write to a non-included namespace without specifying namespace filtering, then the embedded verifier will detect the inconsistency (being the writes to the non-synced namespaces)
- A beta sync with ignorePreExistingNamespaces. In this case mongosync will ignore existing data but error out on any namespace collisions. There is no verification for this case.

I think it would make sense to remove that sentence, because it might just cause confusion.

source/faq.txt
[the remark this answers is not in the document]
mvankeulen94: @rohshar would you mind suggesting text here describing what mongosync does to the destination mid-sync with indexes etc?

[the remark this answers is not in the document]
rohshar: It is actually all of the stuff listed in this page that was deleted: https://github.com/mongodb/docs-cluster-to-cluster-sync/pull/377/files#diff-3217c987a3d0a4d811eb9ed3da436e4576c18570c85a731b618d2178de89e850L34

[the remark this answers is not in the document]
mvankeulen94: Gotcha, thanks @rohshar . Do you mind taking a stab at adding some suggestions to that to-be-deleted text to confirm which pieces we can keep and which we can remove? Any action requested from the user for cleaning up mongosync we can remove, but any explanatory text we can keep.

source/faq.txt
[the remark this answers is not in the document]
mvankeulen94: I'm good to remove that sentence, thanks!

source/about-mongosync.txt
```diff
@@ -157,6 +157,28 @@ If the reverse sync is successful, ``mongosync`` enters the ``RUNNING`` state.
...
+----------------------------------
+
+Consider the following when using ``mongosync`` for continuous synchronization:
+
+- For the destination cluster to have an up to date, consistent view of the 
+  source cluster’s data, ``mongosync`` must :ref:`commit <c2c-api-commit>` . 
+  If the source cluster shuts down before ``mongosync`` can commit, such as in 
+  a disaster scenario, the destination cluster might not have a consistent 
+  snapshot of the source data, and transactions might not appear atomically on 
+  the destination cluster.
```
mvankeulen94, line 172: it would be great if we could add a link here to the content describing all the things that mongosync modifies on the destination (from this [thread](https://github.com/mongodb/docs-cluster-to-cluster-sync/pull/377/files#r1723941662))

source/about-mongosync.txt
```diff
@@ -157,6 +157,28 @@ If the reverse sync is successful, ``mongosync`` enters the ``RUNNING`` state.
...
+  
+  - For any continuous synchronization use cases with ``mongosync``, ensure that 
+    ``mongosync`` commits before cutting over from the source to the 
+    destination.
```
mvankeulen94, line 176: Would be good to also add, "note that there is no way to resume the continuous sync between the two clusters once commit occurs, since mongosync can only sync into empty destination clusters. Use a fresh destination cluster to start a new continuous sync if desired." CC @rohshar

source/reference/mongosync/mongosync-behavior.txt
```diff
@@ -140,16 +140,15 @@ Consistency
 ~~~~~~~~~~~
 
 ``mongosync`` supports :term:`eventual consistency` on the destination
-cluster. Read Consistency is not guaranteed on the destination cluster,
-the source and destination clusters may differ at a given point in
-time. However, if new writes are halted on the source cluster, the
-source and destination clusters will eventually hold the same data.
+cluster. Read consistency is not guaranteed on the destination cluster until 
+commit. Before committing, the source and destination clusters may differ at a 
+given point in time.
```
mvankeulen94, line 145: it would be great to link to the context to-be-added (from this [thread](https://github.com/mongodb/docs-cluster-to-cluster-sync/pull/377/files#r1723941662)) here as well.

source/about-mongosync.txt
```diff
@@ -157,6 +157,28 @@ If the reverse sync is successful, ``mongosync`` enters the ``RUNNING`` state.
...
+Considerations for Continuous Sync
+----------------------------------
+
+Consider the following when using ``mongosync`` for continuous synchronization:
+
+- For the destination cluster to have an up to date, consistent view of the 
+  source cluster’s data, ``mongosync`` must :ref:`commit <c2c-api-commit>` . 
+  If the source cluster shuts down before ``mongosync`` can commit, such as in 
+  a disaster scenario, the destination cluster might not have a consistent 
+  snapshot of the source data, and transactions might not appear atomically on
```
mvankeulen94, line 171: Question for @rohshar , I would like to confirm my understanding of what happens in a hypothetical scenario where mongosync is mid-sync and the source connection severs, assuming the user performs the steps in the disaster recovery tutorial to undo all the temporary changes that mongosync does on the destination.

Aside from torn transactions, are there any other consequences to call out other than just having potentially stale data? E.g., if a user could guarantee they didn't use transactions, what else could go wrong if their source crashed prior to commit and they manually tried to undo mongosync's work?

mvankeulen94 (changes requested):
added some more comments, thank you for bearing with all the changes here!

source/faq.txt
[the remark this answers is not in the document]
ajhuh-mdb: Perfect, re-added the table in the about mongosync page under the continuous sync section and removed the action-driven sentences as suggested.

source/about-mongosync.txt
ajhuh-mdb: Added a note!

source/faq.txt
```diff
@@ -20,20 +20,18 @@ Support.
...
 
-During an ongoing sync, you can write to any non-synced namespaces in the 
-destination cluster as long as the source cluster doesn’t include a namespace 
-with the same name. If you write to a synced namespace before issuing a 
-:ref:`commit <c2c-api-commit>` and while ``canWrite`` is ``false``, the behavior 
-is undefined. To avoid this undefined behavior, you can enable 
-:ref:`write blocking <c2c-dr-write-blocking>`.
+If you write to a synced namespace before issuing a :ref:`commit 
+<c2c-api-commit>` and while ``canWrite`` is ``false``, the behavior is 
+undefined. To avoid this undefined behavior, you can enable :ref:`write blocking
```
rohshar, line 30: I think we should say to ensure that no writes happen to synced namespaces instead of "to avoid this"

source/about-mongosync.txt
```diff
@@ -157,6 +157,73 @@ If the reverse sync is successful, ``mongosync`` enters the ``RUNNING`` state.
...
+Consider the following when using ``mongosync`` for continuous synchronization.
+
+Consistency 
+~~~~~~~~~~~
+
+For the destination cluster to have an up to date, consistent view of the 
+source cluster’s data, ``mongosync`` must :ref:`commit <c2c-api-commit>` . 
+If the source cluster shuts down before ``mongosync`` can commit, such as in 
+a disaster scenario, the destination cluster might not have a consistent 
+snapshot of the source data, and transactions might not appear atomically on
```
rohshar, line 174: I don't think we need to mention transactions again.

I also think we need to be clear that the lack of consistent snapshot is because writes are applied in parallel and mongosync is eventually consistent.

The altering collection/index characteristics below should be in reference to other things that mongosync alters during a sync

source/about-mongosync.txt
```diff
@@ -157,6 +157,73 @@ If the reverse sync is successful, ``mongosync`` enters the ``RUNNING`` state.
...
+restored during the commit process.
+
+.. list-table::
+   :header-rows: 1
+
+   * - Change
+     - Description
+  
+   * - Unique Indexes
+     - Unique indexes on the source cluster synchronize as non-unique indexes
```
rohshar, line 186: are synced? synchronize feels weird to me

source/about-mongosync.txt
```diff
@@ -157,6 +157,73 @@ If the reverse sync is successful, ``mongosync`` enters the ``RUNNING`` state.
...
+For the destination cluster to have an up to date, consistent view of the 
+source cluster’s data, ``mongosync`` must :ref:`commit <c2c-api-commit>` . 
+If the source cluster shuts down before ``mongosync`` can commit, such as in 
+a disaster scenario, the destination cluster might not have a consistent 
+snapshot of the source data, and transactions might not appear atomically on 
+the destination cluster. This can occur because ``mongosync`` temporarily alters 
+some collection characteristics during synchronization. The original values are 
+restored during the commit process.
+
+.. list-table::
```
rohshar, line 179: I think we need a table header here saying something like "Temporarily changed characteristics during a sync" or something

rohshar (changes requested):

source/reference/mongosync/mongosync-behavior.txt
```diff
@@ -140,16 +140,15 @@ Consistency
...
-time. However, if new writes are halted on the source cluster, the
-source and destination clusters will eventually hold the same data.
+cluster. Read consistency is not guaranteed on the destination cluster until 
+commit. Before committing, the source and destination clusters may differ at a 
+given point in time. To learn more, see :ref:`mongosync-considerations`.
 
-Individual writes may be reordered or combined as they are relayed from
-source to destination. For a given document, the total number of writes
-may differ between source and destination.
+While ``mongosync`` is syncing, individual writes may be reordered or combined
```
mvankeulen94, line 147: can we rephrase this sentence to be "mongosync may reorder or combine writes as it relays them from source to destination."

source/about-mongosync.txt
```diff
@@ -157,6 +157,73 @@ If the reverse sync is successful, ``mongosync`` enters the ``RUNNING`` state.
...
+
+Consistency 
+~~~~~~~~~~~
+
+For the destination cluster to have an up to date, consistent view of the 
+source cluster’s data, ``mongosync`` must :ref:`commit <c2c-api-commit>` . 
+If the source cluster shuts down before ``mongosync`` can commit, such as in 
+a disaster scenario, the destination cluster might not have a consistent 
+snapshot of the source data, and transactions might not appear atomically on 
+the destination cluster. This can occur because ``mongosync`` temporarily alters
```
mvankeulen94, line 175: I think we may want to merge the behavior section into here (including the reordering/combination of writes and the note on transactions) and then link to this section on the Behavior page. It'd be great if we could have one place where we state that mongosync, on top of temporarily altering these collection characteristics, also does those other modifications that make the destination not be an accurate representation of the source at any given time.

source/about-mongosync.txt
```diff
@@ -157,6 +157,73 @@ If the reverse sync is successful, ``mongosync`` enters the ``RUNNING`` state.
...
+       on the destination cluster.
+
+   * - TTL Indexes
+     - Synchronization sets ``expireAfterSeconds`` to the value of ``MAX_INT``
+       on the destination cluster.
+
+   * - Hidden Indexes
+     - Synchronization replicates hidden indexes as non-hidden.
+
+   * - Write Blocking
```
mvankeulen94, line 196: I don't follow what mongosync is changing with this write blocking. do we need this note on write blocking here given the other note on write blocking in https://www.mongodb.com/docs/cluster-to-cluster-sync/current/reference/mongosync/mongosync-behavior/#std-label-c2c-write-blocking @rohshar ?

source/about-mongosync.txt
```diff
@@ -157,6 +157,73 @@ If the reverse sync is successful, ``mongosync`` enters the ``RUNNING`` state.
...
+  
+For any continuous synchronization use cases with ``mongosync``, ensure that 
+``mongosync`` commits before cutting over from the source to the 
+destination.
+
+.. note:: 
+  
+   After commit, you can't resume continuous sync between two clusters since 
+   ``mongosync`` can only sync into empty destination clusters. To start a 
+   new continuous sync operation, use a new empty destination cluster.
```
mvankeulen94, line 218: CC @rohshar we may also want to add a note on reverse here.

mvankeulen94 (changes requested):

source/about-mongosync.txt
ajhuh-mdb: Gotcha. In that case, I'll split these up into different sections.

source/about-mongosync.txt
mvankeulen94: You can also use https://www.mongodb.com/docs/cluster-to-cluster-sync/current/reference/api/reverse/ if you'd like to reverse the direction of the sync. This will ensure that any writes that occurred on the destination after cutover are replicated to the original source cluster.

source/about-mongosync.txt
```diff
@@ -157,6 +157,73 @@ If the reverse sync is successful, ``mongosync`` enters the ``RUNNING`` state.
 The synchronization continues in the reverse direction from the original 
 sync job. 
 
+.. _mongosync-considerations:
+
+Considerations for Continuous Sync
```
mvankeulen94, line 162: could we move to behavior page? will see

source/about-mongosync.txt
ajhuh-mdb: discussed in zoom + later comment in this thread that we'll merge this content in the mongosync Behavior page with the migration of the Continuous Sync section to that page.

source/about-mongosync.txt
ajhuh-mdb: ^ at second look, adding the note on filtered sync / reverse limitations will likely be out of scope for this note, but by linking the reverse endpoint, I think we've appropriately covered our bases and left a trail for users to follow to see those limiations.

source/about-mongosync.txt
```diff
@@ -157,6 +157,73 @@ If the reverse sync is successful, ``mongosync`` enters the ``RUNNING`` state.
...
+----------------------------------
+
+Consider the following when using ``mongosync`` for continuous synchronization.
+
+Consistency 
+~~~~~~~~~~~
+
+For the destination cluster to have an up to date, consistent view of the
```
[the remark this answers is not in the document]
ajhuh-mdb: Moved over this text to the mongosync behavior page!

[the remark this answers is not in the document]
ajhuh-mdb: Opted to keep the bullet that's directly in relation to continuous sync/disasters, but added in a sentence that says "To learn more, see [Consistency section]"

ajhuh-mdb:
Thanks both for your patience on this one @mvankeulen94 @rohshar. I believe I've addressed everything we've discussed, so whenever you get the chance, could you re-review? Thanks again.

source/faq.txt
```diff
@@ -20,20 +20,18 @@ Support.
 Can I perform reads or writes to my destination cluster while ``mongosync`` is syncing?
 ---------------------------------------------------------------------------------------
 
-You can perform reads during synchronization if ``canCommit`` is ``true``. 
-However, the data that you read is 
-:term:`eventually consistent <eventual consistency>`, meaning that you might not 
-always read the latest writes.
+You can perform reads during synchronization at any time. However, the data 
+that you read is :term:`eventually consistent <eventual consistency>`, meaning
```
mvankeulen94, line 24: I think the "meaning that you might not always read the latest writes" can be removed altogether, since the consistency link provides the full context on what reads might look like pre-commit.

mvankeulen94 (changes requested):
this is looking great, thanks! just one more comment from me. will share this internally too so I can get approval before we push

source/faq.txt
ajhuh-mdb: That makes sense! Updating.

ajhuh-mdb: EDIT: I'm leaning towards leaving it as is since users might be skimming through this FAQ quickly and might not want to always click into the glossary entry for `eventually consistent` and be routed to a different set of docs pages (the server manual, which is where the glossary lives).

mvankeulen94: thanks! I'd push for removing that text since it's not technically accurate - we've seen folks who have interpreted this to mean that stopping the source cluster mid-sync just means that the destination has a stale snapshot of the source, which isn't the case.

ajhuh-mdb: gotcha! i'll go ahead and remove then.

source/faq.txt
```diff
@@ -20,20 +20,18 @@ Support.
 Can I perform reads or writes to my destination cluster while ``mongosync`` is syncing?
 ---------------------------------------------------------------------------------------
 
-You can perform reads during synchronization if ``canCommit`` is ``true``. 
-However, the data that you read is 
-:term:`eventually consistent <eventual consistency>`, meaning that you might not 
-always read the latest writes.
+You can perform reads during synchronization at any time. However, the data 
+that you read is :term:`eventually consistent <eventual consistency>`. For 
+up-to-date reads, wait for the migration to commit. To learn more, see
```
mvankeulen94, line 25: apologies that I missed this earlier, we should change "up-to-date" with "consistent" here as well, for the same reasons as the [prior thread](https://github.com/mongodb/docs-cluster-to-cluster-sync/pull/377#discussion_r1755426401) on the 'latest writes' verbiage that we're deleting.
no rush to update the staging on this new change BTW, I've shared the staging env with the team and I've mentioned this note above as well.

rohshar (approved):

mvankeulen94 (approved):
lgtm, thanks!

PR #380, opened 2024-08-12
DOCSP-37061-mongosync-timestamp-limitation

ajhuh-mdb:
## DESCRIPTION
- Add timestamp limitation 

## STAGING 
https://preview-mongodbajhuhmdb.gatsbyjs.io/cluster-sync/DOCSP-37061-mongosync-timestamp-limitation/reference/limitations/#general-limitations

## JIRA 
https://jira.mongodb.org/browse/DOCSP-37061

## BUILD 
https://workerpool-boxgs.mongodbstitch.com/pages/job.html?collName=queue&jobId=66ba647a83839571177a6a44

### DIFF
--- a/source/reference/limitations.txt
+++ b/source/reference/limitations.txt
@@ -61,8 +61,8 @@ General Limitations
   <https://www.mongodb.com/docs/atlas/atlas-search/create-index/>`__.
 - ``mongosync`` only supports clusters that use the :ref:`WiredTiger 
   <storage-wiredtiger>` storage engine.
-- ``mongosync`` does not support migrating documents that have an empty 
-  top-level timestamp, such as ``Timestamp(0,0)``.
+- You can't sync a collection with any documents that have an empty timestamp,
+  such as ``Timestamp(0,0)``. 
 
 MongoDB Community Edition
 -------------------------
--- a/source/reference/limitations.txt
+++ b/source/reference/limitations.txt
@@ -61,6 +61,8 @@ General Limitations
   <https://www.mongodb.com/docs/atlas/atlas-search/create-index/>`__.
 - ``mongosync`` only supports clusters that use the :ref:`WiredTiger 
   <storage-wiredtiger>` storage engine.
+- ``mongosync`` does not support migrating documents that have an empty 
+  top-level timestamp, ``Timestamp(0,0)``.
 
 MongoDB Community Edition
 -------------------------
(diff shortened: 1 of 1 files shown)


source/reference/limitations.txt
```diff
@@ -61,6 +61,8 @@ General Limitations
   <https://www.mongodb.com/docs/atlas/atlas-search/create-index/>`__.
 - ``mongosync`` only supports clusters that use the :ref:`WiredTiger 
   <storage-wiredtiger>` storage engine.
+- ``mongosync`` does not support migrating documents that have an empty 
+  top-level timestamp, such as ``Timestamp(0,0)``.
```
ianf-mongodb, line 65: **Question** 
When you say not supported, what does this mean? Will the documents in a collection be skipped if they contain these types of documents? Will the sync completely fail? I think we should elaborate on the end result here.

Also I think adding a small example document with a json code block directive might better illustrate the types of documents that are being referenced here. 

Something like:

```
For example, the following document is not support because it has an empty timestamp field:
.. code-block:: json

   {
      "ts" : "",
      "a" : "1",
      "b" : "2"
   }

ianf-mongodb (commented):
Thanks @ajhuh-mdb suggestion and question regarding the update in this edit.

source/reference/limitations.txt
ajhuh-mdb: The behavior for this is undefined  — I can explicitly mention this.

ajhuh-mdb: As a follow-up to this, the top of the limitations page mentions that running `mongosync` with any of the listed limitations could lead to undefined behavior, so to avoid redundancy, I'm leaning towards leaving this as is.

ajhuh-mdb: As for your second comment, I adjusted the wording to mention that _any_ document that contains an empty timestamp isn't supported. I took the copy from the [cloud docs](https://github.com/10gen/cloud-docs/pull/5360/files#diff-ac33dbc61c8f5175ac61665d5cb9f78393f0ae0254817a804ee4b215f2883572R38) regarding the same limitation.

ajhuh-mdb:
Thanks for taking a look @ianf-mongodb! I left some responses to your comments in the threads above.

source/reference/limitations.txt
ianf-mongodb: Awesome, I think this is much clearer for the behavior.

ianf-mongodb (approved):
Thanks @ajhuh-mdb this LGTM

jddubois (approved):

rohshar (approved):

PR #404, opened 2024-08-23
DOCSP-42834-beta-access-not-guaranteed

ajhuh-mdb:
## DESCRIPTION
Add note to mention that access to beta binaries isn't guaranteed with a support ticket.

## STAGING 
https://preview-mongodbajhuhmdb.gatsbyjs.io/cluster-sync/DOCSP-42834-beta-access-not-guaranteed/reference/beta-program/#get-started

## JIRA
https://jira.mongodb.org/browse/DOCSP-42834

## BUILD
https://workerpool-boxgs.mongodbstitch.com/pages/job.html?collName=queue&jobId=66c89c3d25dc1b90c748b7d3


nvillahermosa-mdb (approved):
LGTM

source/reference/beta-program.txt
```diff
@@ -25,6 +25,11 @@ Get Started
 To request access to the {+c2c-beta-program-short+} binaries, please file a 
 ticket with `support <https://support.mongodb.com/>`_.
 
+.. note:: 
+
+   Filing a support ticket does  *not* guarantee access to the 
+   {+c2c-beta-program-short+} binaries.
```
mvankeulen94, line 31: Can we qualify further to be "To request access to the mongosync beta binaries, please review the below disclaimer. Note that requests that do not conform to this disclaimer will **not** be approved. To continue with the request, please file a ticket with support."?

mvankeulen94 (changes requested):
thanks! left a comment

source/reference/beta-program.txt
ajhuh-mdb: I left out the mention of the disclaimer because in copy, we mention that users must accept the disclaimer when running the binary for the first time (which implies they already have access to the binary). In that case, should we remove the verbiage about the disclaimer popping up when running the binary?

mvankeulen94: We have seen cases where customer hadn't reviewed the disclaimer before submitting the access request. My hope here is to 1) ask that they review the disclaimer before filing a ticket 2) also be aware that they need to explicitly accept the terms of the disclaimer when using the binary

mvankeulen94 (changes requested):

source/reference/beta-program.txt
ajhuh-mdb: gotcha. updated

source/reference/beta-program.txt
```diff
@@ -49,7 +52,8 @@ following disclaimer:
    migrations). Beta capabilities are not intended for use to migrate 
    production workloads.
 
-   Do you want to continue? (y/n):
+When you first run the {+c2c-beta-program-short+} binary, you must accept the 
+disclaimer again.
```
mvankeulen94, line 56: nit: we can adjust "you must [...]" to be "you will be prompted to accept the disclaimer". I think the "again" can be confusing, since the mongosync binary has a command line prompt that requires the user to accept via the prompt, which would only happen once they have access to the binary

mvankeulen94 (commented):
just a small nit, then am good to approve

mvankeulen94 (approved):
lgtm, thank you!

PR #408, opened 2024-09-10
DOCSP-43294 Cutover Clarifications

kennethdyer:
## Description

Clarifies the cutover process.

## Staging
* [Finalize Cutover](https://deploy-preview-408--docs-cluster-to-cluster-sync.netlify.app/reference/cutover-process/)
* [commit intro](https://deploy-preview-408--docs-cluster-to-cluster-sync.netlify.app/reference/api/commit/) 

## JIRA

[DOCSP-43294](https://jira.mongodb.org/browse/DOCSP-43294)

### DIFF
--- a/source/reference/cutover-process.txt
+++ b/source/reference/cutover-process.txt
@@ -178,16 +178,15 @@ canWrite and COMMITTED
 ~~~~~~~~~~~~~~~~~~~~~~
 
 ``mongosync`` permits writes on the destination cluster at much earlier stage than
-the ``COMMITTED`` state. While you can write to the destination cluster after it
-reports ``canWrite: true``, you may encounter unexpected results until ``mongosync``
-reaches the ``COMMITTED`` state.
+the ``COMMITTED`` state.
 
-- In the initial sync, ``mongosync`` replicates unique indexes on the source
-  cluster as non-unique indexes on the destination cluster. When writes are possible
-  on the destination cluster, ``mongosync`` begins to convert these non-unique indexes
-  to unique indexes.
+In the initial sync, ``mongosync`` replicates unique indexes on the source
+cluster as non-unique indexes on the destination cluster. When writes are possible
+on the destination cluster, ``mongosync`` begins to convert these non-unique indexes
+to unique indexes. Queries on the destination cluster that require a unique
+index may return unexpected results.
 
-  This process may take a long time when syncing large collections.
+This process may take a long time when syncing large collections.
 
 Learn More
 ----------
--- a/source/reference/cutover-process.txt
+++ b/source/reference/cutover-process.txt
@@ -117,6 +117,7 @@ Steps
       shows ``canWrite`` is ``true``.
 
       .. io-code-block::
+         :copyable: true
 
          .. input::
             :language: sh
--- a/source/reference/cutover-process.txt
+++ b/source/reference/cutover-process.txt
@@ -183,10 +183,11 @@ the ``COMMITTED`` state.
 In the initial sync, ``mongosync`` replicates unique indexes on the source
 cluster as non-unique indexes on the destination cluster. When writes are possible
 on the destination cluster, ``mongosync`` begins to convert these non-unique indexes
-to unique indexes. Queries on the destination cluster that require a unique
-index may return unexpected results.
+to unique indexes.
 
-This process may take a long time when syncing large collections.
+This process may take a long time when syncing large collections. Queries
+on the destination cluster that require a unique index may return unexpected
+results.
 
 Learn More
 ----------
--- a/source/reference/cutover-process.txt
+++ b/source/reference/cutover-process.txt
@@ -180,15 +180,17 @@ canWrite and COMMITTED
 ``mongosync`` permits writes on the destination cluster at an
 earlier stage than the ``COMMITTED`` state.
 
-In the initial sync, ``mongosync`` replicates unique indexes on
-the source cluster as non-unique indexes on the destination
-cluster. When the ``/progress`` endpoint sets ``canWrite`` to
-``true``, ``mongosync`` converts these non-unique indexes into
-unique indexes.
+During the initial sync, ``mongosync`` replicates unique indexes
+on the source cluster as non-unique indexes on the destination
+cluster. During the commit, the relevant indexes on the
+destination cluster are set to ``prepareUnique``. When this is
+done, the ``/progress`` endpoint begins to return ``canWrite:
+true``. ``mongosync`` then converts the ``prepareUnique``
+indexes into unique indexes. When this is done, ``mongosync``
+changes its state to ``COMMITTED``.
 
 This process may take a long time when syncing large
-collections. Queries on the destination cluster that require a
-unique index may return unexpected results.
+collections.
 
 Learn More
 ----------
--- a/source/reference/cutover-process.txt
+++ b/source/reference/cutover-process.txt
@@ -168,17 +168,19 @@ Steps
       .. code-block:: javascript
 
          db.adminCommand(
-           {
-             setUserWriteBlockMode: 1,
-             global: false
-           }
+            {
+               setUserWriteBlockMode: 1,
+               global: false
+            }
          )
+
+      Then, transfer your application workload to the destination cluster.
 
    .. step:: Call the ``progress`` endpoint to determine the status of the ``mongosync`` process.
 
-      When the ``mongosync`` progress response indicates that the 
+      When the ``mongosync`` progress response indicates that the
       ``mongosync`` state is ``COMMITTED``, the cutover process is
-      complete. 
+      complete.
 
       Request
       ~~~~~~~
--- a/source/reference/cutover-process.txt
+++ b/source/reference/cutover-process.txt
@@ -198,40 +198,17 @@ Steps
       ``mongosync`` state is ``COMMITTED``, the cutover process is
       complete.
 
-      Request
-      ~~~~~~~
+      .. io-code-block::
 
-      .. literalinclude:: /includes/api/requests/progress.sh
-         :language: shell
+         .. input::
+            :language: sh
 
-      Response
-      ~~~~~~~~
+            curl localhost:27182/api/v1/progress -XGET | jq .progress.state
 
-      .. code-block:: json
-         :emphasize-lines: 4
-         :copyable: false
+         .. output::
+            :language: json
 
-         {
-            "progress":
-               {
-                  "state":"COMMITTED",
-                  "canCommit":true,
-                  "canWrite":false,
-                  "info":"change event application",
-                  "lagTimeSeconds":0,
-                  "collectionCopy":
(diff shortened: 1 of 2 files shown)


source/reference/cutover-process.txt
```diff
@@ -164,54 +142,52 @@ Steps
...
-                     },
-                  "directionMapping":
-                     {
-                        "Source":"cluster0: localhost:27017",
-                        "Destination":"cluster1: localhost:27018"
-                     }
-               },
-            "success": true
-         }
+``mongosync`` permits writes on the destination cluster at much earlier stage than
```
jason-price-mongodb, line 180: Consider "at an earlier"

source/reference/cutover-process.txt
```diff
@@ -164,54 +142,52 @@ Steps
...
+the ``COMMITTED`` state.
 
+In the initial sync, ``mongosync`` replicates unique indexes on the source
+cluster as non-unique indexes on the destination cluster. When writes are possible
+on the destination cluster, ``mongosync`` begins to convert these non-unique indexes
```
jason-price-mongodb, line 185: Perhaps just "mongosync converts these"

source/reference/cutover-process.txt
```diff
@@ -164,54 +142,52 @@ Steps
...
+to unique indexes.
 
+This process may take a long time when syncing large collections. Queries
```
jason-price-mongodb, line 188: If poss, maybe add an estimate or ballpark on the time. 1 h, 6 h, ...

source/reference/cutover-process.txt
```diff
@@ -164,54 +142,52 @@ Steps
...
+on the destination cluster that require a unique index may return unexpected
+results.
```
jason-price-mongodb, line 190: If poss, add some info on what would be unexpected.

jason-price-mongodb (approved):
@kennethdyer LGTM with suggestions. Cheers.

source/reference/cutover-process.txt
kennethdyer: Fixed

source/reference/cutover-process.txt
kennethdyer: I suspect it varies widely based on size and complexity of indexes.  Will leave this comment up for the external reviewer to weight in.

source/reference/cutover-process.txt
kennethdyer: Leaving this up as well for the external.

source/reference/cutover-process.txt
```diff
@@ -164,54 +142,52 @@ Steps
...
+         :copyable: true
 
-      .. literalinclude:: /includes/api/requests/progress.sh
-         :language: shell
+         .. input::
+            :language: bash
 
-      Response
-      ~~~~~~~~
+            curl localhost:27182/api/v1/progress -XGET | jq ".progress.state"
```
tdq45gj, line 165: Same as above.

source/reference/cutover-process.txt
```diff
@@ -114,41 +115,18 @@ Steps
...
-
-      Response
-      ~~~~~~~~
+         .. input::
+            :language: bash
 
-      .. code-block:: json
-         :emphasize-lines: 6
-         :copyable: false
+            curl localhost:27182/api/v1/progress -XGET | jq ".progress.canWrite"
```
tdq45gj, line 124: I wonder if it's better to add `-sS` flag so that `curl` doesn't print the progress bar.

source/reference/cutover-process.txt
tdq45gj: I don't think we have an estimate of how long it takes. I agree that it's workload dependent.

source/reference/cutover-process.txt
tdq45gj: That's a good question. I don't think anything would be unexpected for unconverted unique indexes because the documents should already be in the destination.

tdq45gj (commented):
Thanks for the changes! I'm requesting another review from @ajayvijayakumar123, especially for the comments on commit behavior.

source/reference/cutover-process.txt
ajayvijayakumar123: ^ agreed. I'm realizing now that I should've provided a bit more context on the JIRA ticket about this section.

I think we should note that when monogsync indicates canWrite:true, mongosync would've already converted these non-unique indexes to have the `prepareUnique` property. This means that you cannot insert _new_ documents that violate unique index constraints on the destination.

Converting `prepareUnique` to `unique` means that the server performs an index scan to see if any preexisting documents violate unique constraints.

These (unique index [docs](https://www.mongodb.com/docs/manual/core/index-unique/convert-to-unique/#std-label-index-convert-to-unique)) and collMod [docs](https://www.mongodb.com/docs/manual/reference/command/collMod/#mongodb-collflag-index) have more information. It might be helpful to link them here.

source/reference/cutover-process.txt
```diff
@@ -164,54 +142,52 @@ Steps
...
-                     }
-               },
-            "success": true
-         }
+``mongosync`` permits writes on the destination cluster at an earlier stage than
+the ``COMMITTED`` state.
 
+In the initial sync, ``mongosync`` replicates unique indexes on the source
+cluster as non-unique indexes on the destination cluster. When ``mongosync``
+re-enables writes on the destination cluster, ``mongosync`` converts these
```
ajayvijayakumar123, line 185: Instead of "re-enables writes", I think we should say "when the progress endpoint indicates `canWrite` is true". This is more consistent with what we say in step 4 above.

ajayvijayakumar123 (changes requested):
Some suggestions!

source/reference/cutover-process.txt
kennethdyer: Fixed

source/reference/cutover-process.txt
kennethdyer: Fixed

source/reference/cutover-process.txt
kennethdyer: Fixed

source/reference/cutover-process.txt
kennethdyer: @ajayvijayakumar123, So the process we should convey:

* During initial sync, unique indexes become non-unique indexes.
* In that process (or after?), the relevant indexes on the destination cluster are set to prepareUnique.
* `/progress` returns `canWrite: true`.
* `mongosync` converts prepareUnique indexes to unique indexes.

Questions:
* If the index scan and conversion to unique raises an error, where would the user look for it?
* If `canWrite: true` happens after the `prepareUnique` value is set, would this mean that the errors originate in duplicates synced from the source cluster?

source/reference/cutover-process.txt
```diff
@@ -164,54 +142,53 @@ Steps
...
+earlier stage than the ``COMMITTED`` state.
 
+In the initial sync, ``mongosync`` replicates unique indexes on
+the source cluster as non-unique indexes on the destination
+cluster. When the ``/progress`` endpoint sets ``canWrite`` to
+``true``, ``mongosync`` converts these non-unique indexes into
+unique indexes.
 
+This process may take a long time when syncing large
+collections. Queries on the destination cluster that require a
```
jddubois, line 190: > Queries on the destination cluster that require a unique index may return unexpected results.

Why is this?

source/reference/cutover-process.txt
jddubois: > If `canWrite: true` happens after the `prepareUnique` value is set, would this mean that the errors originate in duplicates synced from the source cluster?

Yes, if there is an error after `canWrite: true`, the duplicates came from the source cluster, assuming that mongosync does not have a bug. In this case, the destination cluster will function identically to the source cluster.

I would recommend we say something like:

"If the commit fails after `canWrite: true`, this means that the source cluster has documents that violate unique index constraints. The destination cluster will not accept any new duplicate values for the failed index constraints, but cannot fully add the constraint due to the duplicates that it has received from the source. If you wish to resolve these duplicates and fully enable the index constraint, manual intervention is required."

ajayvijayakumar123: Apologies for the late response.

The process would:
* During initial sync, unique indexes become non-unique indexes
* _During commit_, the relevant indexes on the destination cluster are set to prepareUnique.
* /progress returns canWrite: true.
* mongosync converts prepareUnique indexes to unique indexes

Good questions. We currently have an engineering proposal (https://jira.mongodb.org/browse/WRITING-25602) on our radar to get consensus on how we're treating these potential errors between canWrite:true and COMMITTED. We want to only provide context on this once the EP is approved. Until then, I think we should only document the differences between canWrite:true and COMMITTED and not the errors that can arise between them. Does that make sense?

kennethdyer: Fixed

source/reference/cutover-process.txt
kennethdyer: Mitigated

kennethdyer:
@ajayvijayakumar123, @tdq45gj, @jddubois, Changes applied. Anything else you all would like to see?

source/reference/cutover-process.txt
```diff
@@ -164,54 +142,57 @@ Steps
...
-               },
-            "success": true
-         }
+``mongosync`` permits writes on the destination cluster at an
+earlier stage than the ``COMMITTED`` state.
 
+In the initial sync, ``mongosync`` replicates unique indexes on
+the source cluster as non-unique indexes on the destination
+cluster. During commit, the relevant non-unique indexes on the
+destination cluster are set to ``prepareUnique``. When this is
```
ajayvijayakumar123, line 186: I think we should note that `prepareUnique` indexes still reject new documents that violate unique index constraints. That way, users can trust that their destination prepareUnique indexes are similar to unique ones.

ajayvijayakumar123 (changes requested):
Thanks for the update! Had 2 more suggestions

tdq45gj (approved):
LGTM, thanks!

source/reference/cutover-process.txt
```diff
@@ -164,54 +142,57 @@ Steps
...
+done, the ``/progress`` endpoint begins to return ``canWrite:
+true``. ``mongosync`` then converts the ``prepareUnique``
+indexes into unique indexes. When this is done, ``mongosync``
+changes its state to ``COMMITTED``.
 
+This process can be resource intensive when syncing large
```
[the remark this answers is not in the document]
kennethdyer: Fixed

source/reference/cutover-process.txt
kennethdyer: Fixed

kennethdyer:
@ajayvijayakumar123, @jddubois, Changes applied.  Anything else you all would like to see?

ajayvijayakumar123 (approved):
LGTM!

jddubois (approved):

PR #505, opened 2024-11-19
DOCSP-45350-mongosync-limitations-always-in-effect

amalhotra-mdb:
## DESCRIPTION
Add an `important` admonition that clarifies that the limitations are still in effect even after pausing/stopping mongosync.

## STAGING
https://deploy-preview-505--docs-cluster-to-cluster-sync.netlify.app/reference/limitations/#limitations

## JIRA
https://jira.mongodb.org/browse/DOCSP-45350

## SELF-REVIEW CHECKLIST

- [ ] Does each file have 3-5 taxonomy facet tags?
- [ ] Is this free of any warnings or errors in the RST?
- [ ] Is this free of spelling errors?
- [ ] Is this free of grammatical errors?
- [ ] Is this free of staging / rendering issues?
- [ ] Are all the links working?

### DIFF
--- a/source/reference/limitations.txt
+++ b/source/reference/limitations.txt
@@ -21,6 +21,10 @@ Limitations
    these limitations could lead to undefined behavior on the destination
    cluster.
 
+   You must adhere to these limitations for the full duration of the the
+   migration, including when the migration is paused or stopped and will be
+   resumed.
+
 General Limitations
 -------------------
 
@@ -63,11 +67,6 @@ General Limitations
   <storage-wiredtiger>` storage engine.
 - You can't sync a collection with any documents that have an empty timestamp,
   such as ``Timestamp(0,0)`` on pre-6.0 source clusters. 
-
-.. important::
-
-   ``mongosync`` enforces the limitations, regardless of whether
-   ``mongosync`` is running, paused, or stopped.
 
 MongoDB Community Edition
 -------------------------
--- a/source/reference/limitations.txt
+++ b/source/reference/limitations.txt
@@ -66,7 +66,7 @@ General Limitations
 
 .. important::
 
-   ``mongosync`` enforces the documented limitations, regardless of whether
+   ``mongosync`` enforces the limitations, regardless of whether
    ``mongosync`` is running, paused, or stopped.
 
 MongoDB Community Edition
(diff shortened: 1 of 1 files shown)


source/reference/limitations.txt
```diff
@@ -64,6 +64,11 @@ General Limitations
 - You can't sync a collection with any documents that have an empty timestamp,
   such as ``Timestamp(0,0)`` on pre-6.0 source clusters. 
 
+.. important::
+
+   ``mongosync`` enforces the documented limitations, regardless of whether
```
kennethdyer, line 69: Succinctness:

```
``mongosync`` enforces the limitations...
```

The "documented limitations" implies undocumented limitations.

kennethdyer (approved):
Thanks @amalhotra-mdb, this LGTM with one suggestion.

source/reference/limitations.txt
amalhotra-mdb: Ah that makes sense. Thank you for this fix!

rohshar (changes requested):
Why is this not listed in the top level warning? I think it would be better there. 

Also, mongosync doesn't enforce all limitations, we can't say that. Just that they have to adhere to the limitaitons for the full duration of the migration, including when the migration is paused / or stopped if it may be resumed

amalhotra-mdb:
> Why is this not listed in the top level warning? I think it would be better there.
> 
> Also, mongosync doesn't enforce all limitations, we can't say that. Just that they have to adhere to the limitaitons for the full duration of the migration, including when the migration is paused / or stopped if it may be resumed

Thanks for your review, @rohshar. I agree, this information should b e in the top level warning – I went ahead and added it there. Thanks for your content changes – all changes made! Could you take a second look and see if it looks okay to you?

source/reference/limitations.txt
```diff
@@ -21,6 +21,10 @@ Limitations
    these limitations could lead to undefined behavior on the destination
    cluster.
 
+   You must adhere to these limitations for the full duration of the the
+   migration, including when the migration is paused or stopped and will be
```
rohshar, line 25:
```suggestion
   migration, including when the migration is paused or stopped if it will be
```

rohshar (approved):
one nit, then LGTM

PR #522, opened 2024-12-11
DOCSP-45860-older-version-limitations

ajhuh-mdb:
## DESCRIPTION 
- Documents limits of Older Version Support in C2C

## STAGING 
https://deploy-preview-522--docs-cluster-to-cluster-sync.netlify.app/reference/limitations/#pre-6.0-migrations

## JIRA
[DOCSP-45860](https://jira.mongodb.org/browse/DOCSP-45860)

### DIFF
--- a/source/includes/fact-older-version-limitations.rst
+++ b/source/includes/fact-older-version-limitations.rst
@@ -1,12 +1,13 @@
 - Writes that produce DDL events cannot occur on the source cluster during the 
   migration. The following events cannot occur: 
   
+  - ``collMod``
   - ``create``
+  - ``createIndexes``
   - ``drop``
   - ``dropDatabase``
-  - ``createIndexes``
   - ``dropIndexes``
-  - ``collModrefineCollectionShardKey``
+  - ``refineCollectionShardKey``  
   - ``rename``
   - ``reshardCollection``
   - ``shardCollection``
@@ -23,14 +24,14 @@
 
 - ``geoHaystack`` indexes are not supported.
 
-- :ref:`/reverse <c2c-api-reverse>` endpoints are not supported. You can't 
+- :ref:`/reverse <c2c-api-reverse>` endpoint is not supported. You can't 
   enable the ``reversible`` option in the :ref:`/start <c2c-api-start>` request.
 
 - You can't enable the ``enableUserWriteBlocking`` option in the ``/start`` 
   request. 
 
 - You can't enable the ``createSupportingIndexes`` :ref:`sharding parameter 
-  <c2c-api-start-sharding>` with a pre-6.0 cluster. 
+  <c2c-api-start-sharding>`. 
 
 - Indexes with inconsistent specs or that are missing on some shards are not 
   supported. To check for index inconsistencies, see 
--- a/source/includes/fact-older-version-limitations.rst
+++ b/source/includes/fact-older-version-limitations.rst
@@ -21,7 +21,7 @@
      Writes that produce DDL events on source collections outside of the 
      :ref:`namespace filter <c2c-filtered-sync>` are allowed.
 
-- geoHaystack indexes <index-geohaystack-index>` are not supported.
+- ``geoHaystack`` indexes are not supported.
 
 - :ref:`/reverse <c2c-api-reverse>` endpoints are not supported. You can't 
   enable ``reversible`` in the :ref:`/start <c2c-api-start>` request.
--- a/source/includes/fact-older-version-limitations.rst
+++ b/source/includes/fact-older-version-limitations.rst
@@ -1,5 +1,6 @@
-- Writes that produce DDL events cannot occur on the source cluster during the 
-  migration. The following events cannot occur: 
+- Writes that produce :term:`DDL <DDL (Data Definition Language)>` events cannot 
+  occur on the source cluster during the migration. The following events cannot 
+  occur: 
   
   - ``collMod``
   - ``create``
@@ -28,13 +29,15 @@
   enable the ``reversible`` option in the :ref:`/start <c2c-api-start>` request.
 
 - You can't enable the ``enableUserWriteBlocking`` option in the ``/start`` 
-  request. 
+  request. Ensure that no writes are made to the source or destination cluster 
+  during the migration.
 
 - You can't enable the ``createSupportingIndexes`` :ref:`sharding parameter 
-  <c2c-api-start-sharding>`. 
+  <c2c-api-start-sharding>`. Instead, create an index to support your shard key 
+  on the source cluster. 
 
-- Indexes with inconsistent specs or that are missing on some shards are not 
-  supported. To check for index inconsistencies, see 
+- If there are any indexes with inconsistent specs or that are missing 
+  ``mongosync`` returns an error. To check for index inconsistencies, see 
   :ref:`manage-indexes-find-inconsistent-indexes`.
 
 - For source clusters running MongoDB 4.4, :ref:`SRV connection strings 
--- a/source/includes/fact-older-version-limitations.rst
+++ b/source/includes/fact-older-version-limitations.rst
@@ -1,0 +1,40 @@
+- Writes that produce DDL events cannot occur on the source cluster during the 
+  migration. The following events cannot occur: 
+  
+  - create
+  - drop
+  - dropDatabase
+  - createIndexes
+  - dropIndexes
+  - collModrefineCollectionShardKey
+  - rename
+  - reshardCollection
+  - shardCollection
+
+  This includes operations that may create new collections such as 
+  :dbcommand:`mapReduce`, :pipeline:`$out`, and :pipeline:`$merge`. This also 
+  includes collections created implicitly from inserts. Only writes that produce 
+  CRUD events can occur during the migration.
+
+  .. note:: 
+   
+     Writes that produce DDL events on source collections outside of the 
+     :ref:`namespace filter <c2c-filtered-sync>` are allowed.
+
+- :ref:`GeoHaystack indexes <index-geohaystack-index>` are not supported.
+
+- :ref:`/reverse <c2c-api-reverse>` endpoints are not supported. You can't 
+  enable ``reversible`` in the :ref:`/start <c2c-api-start>` request.
+
+- You can't enable the ``enableUserWriteBlocking`` in the ``/start`` request. 
+
+- :ref:`SRV connection strings <connections-dns-seedlist>` for 4.4 source 
+  clusters are not supported. You must use a :ref:`standard connection string
+  <connections-standard-connection-string-format>`.
+
+- You can't enable the ``createSupportingIndexes`` :ref:`sharding parameter 
+  <c2c-api-start-sharding>` with a pre-6.0 cluster. 
+
+- Indexes with inconsistent specs or are missing on some shards are not 
+  supported. To check for index inconsistencies, see 
+  :ref:`manage-indexes-find-inconsistent-indexes`.
--- a/source/includes/fact-older-version-limitations.rst
+++ b/source/includes/fact-older-version-limitations.rst
@@ -1,15 +1,15 @@
 - Writes that produce DDL events cannot occur on the source cluster during the 
   migration. The following events cannot occur: 
   
-  - create
-  - drop
-  - dropDatabase
-  - createIndexes
-  - dropIndexes
-  - collModrefineCollectionShardKey
-  - rename
-  - reshardCollection
-  - shardCollection
+  - ``create``
+  - ``drop``
+  - ``dropDatabase``
+  - ``createIndexes``
+  - ``dropIndexes``
+  - ``collModrefineCollectionShardKey``
+  - ``rename``
+  - ``reshardCollection``
+  - ``shardCollection``
 
   This includes operations that may create new collections such as 
   :dbcommand:`mapReduce`, :pipeline:`$out`, and :pipeline:`$merge`. This also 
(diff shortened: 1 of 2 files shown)


source/includes/fact-older-version-limitations.rst
```diff
@@ -0,0 +1,40 @@
...
+  
+  - create
+  - drop
+  - dropDatabase
+  - createIndexes
+  - dropIndexes
+  - collModrefineCollectionShardKey
+  - rename
+  - reshardCollection
+  - shardCollection
```
jeff-allen-mongo, line 12: [question]

Should we monospace these?

source/includes/fact-older-version-limitations.rst
```diff
@@ -0,0 +1,40 @@
...
+  :dbcommand:`mapReduce`, :pipeline:`$out`, and :pipeline:`$merge`. This also 
+  includes collections created implicitly from inserts. Only writes that produce 
+  CRUD events can occur during the migration.
+
+  .. note:: 
+   
+     Writes that produce DDL events on source collections outside of the 
+     :ref:`namespace filter <c2c-filtered-sync>` are allowed.
+
+- :ref:`GeoHaystack indexes <index-geohaystack-index>` are not supported.
```
jeff-allen-mongo, line 24: [Question]

Do we need to mention this? These have been removed since 5.0 and we don't have any docs on them anymore.

I think if we're keeping this, we want `geoHaystack` to be in camel case with a lower-case `g`.

source/includes/fact-older-version-limitations.rst
```diff
@@ -0,0 +1,40 @@
...
+
+- :ref:`/reverse <c2c-api-reverse>` endpoints are not supported. You can't 
+  enable ``reversible`` in the :ref:`/start <c2c-api-start>` request.
+
+- You can't enable the ``enableUserWriteBlocking`` in the ``/start`` request.
```
jeff-allen-mongo, line 29:
```suggestion
- You can't enable ``enableUserWriteBlocking`` in the ``/start`` request. 
```

jeff-allen-mongo: Or maybe "enable the `enableUserWriteBlocking` option". Whichever you prefer.

source/includes/fact-older-version-limitations.rst
```diff
@@ -0,0 +1,40 @@
...
+     :ref:`namespace filter <c2c-filtered-sync>` are allowed.
+
+- :ref:`GeoHaystack indexes <index-geohaystack-index>` are not supported.
+
+- :ref:`/reverse <c2c-api-reverse>` endpoints are not supported. You can't 
+  enable ``reversible`` in the :ref:`/start <c2c-api-start>` request.
+
+- You can't enable the ``enableUserWriteBlocking`` in the ``/start`` request. 
+
+- :ref:`SRV connection strings <connections-dns-seedlist>` for 4.4 source
```
jeff-allen-mongo, line 31: Making sure I'm understanding this correctly - this limitation only applies to 4.4 source clusters, not all pre-6.0 clusters?

source/includes/fact-older-version-limitations.rst
```diff
@@ -0,0 +1,40 @@
...
+- You can't enable the ``enableUserWriteBlocking`` in the ``/start`` request. 
+
+- :ref:`SRV connection strings <connections-dns-seedlist>` for 4.4 source 
+  clusters are not supported. You must use a :ref:`standard connection string
+  <connections-standard-connection-string-format>`.
+
+- You can't enable the ``createSupportingIndexes`` :ref:`sharding parameter 
+  <c2c-api-start-sharding>` with a pre-6.0 cluster. 
+
+- Indexes with inconsistent specs or are missing on some shards are not
```
jeff-allen-mongo, line 38:
```suggestion
- Indexes with inconsistent specs or that are missing on some shards are not 
```

jeff-allen-mongo (changes requested):
@ajhuh-mdb thanks for these updates! Looks good so far - just have some comments and suggestions for your consideration.

source/includes/fact-older-version-limitations.rst
ajhuh-mdb: ah yes we should — overlooked that. updating!

source/includes/fact-older-version-limitations.rst
ajhuh-mdb: fixed/updated

source/includes/fact-older-version-limitations.rst
ajhuh-mdb: @ajayvijayakumar123 can you confirm on whether this is true?

source/includes/fact-older-version-limitations.rst
ajhuh-mdb: Fixed

source/includes/fact-older-version-limitations.rst
ajayvijayakumar123: Yup I agree with Ali. Also I think we still have some docs on them: https://www.mongodb.com/docs/manual/core/indexes/index-types/geospatial/geohaystack/

source/includes/fact-older-version-limitations.rst
ajayvijayakumar123: yup just for 4.4 sources

jeff-allen-mongo: Maybe we can reorder the phrases here slightly so it's clearer on a quick scan that this only applies to 4.4. Like:

"For 4.4 source clusters, [SRV connection strings](https://www.mongodb.com/docs/master/reference/connection-string/#std-label-connections-dns-seedlist) are not supported. You must use a [standard connection string.](https://www.mongodb.com/docs/master/reference/connection-string/#std-label-connections-standard-connection-string-format)"

I'd also maybe consider moving this to the end of the list since it's kind of in its own category of only applying to 4.4 clusters.

source/includes/fact-older-version-limitations.rst
ajayvijayakumar123: yup that works!

source/includes/fact-older-version-limitations.rst
```diff
@@ -0,0 +1,42 @@
...
+  
+  - ``create``
+  - ``drop``
+  - ``dropDatabase``
+  - ``createIndexes``
+  - ``dropIndexes``
+  - ``collModrefineCollectionShardKey``
+  - ``rename``
+  - ``reshardCollection``
+  - ``shardCollection``
```
jeff-allen-mongo, line 12: [Suggestion]

Maybe alphabetize?

jeff-allen-mongo (approved):
LGTM % a final suggestion. Thanks @ajhuh-mdb!

source/includes/fact-older-version-limitations.rst
ajhuh-mdb: I'm inclined to leave this as is, or at least with few swaps, because I felt it'd be good to order it this way to users see the most familiar events at the top. If we swapped this to alphabetical, I though it could be a bit jarring for users to first see ``collModRefineCollectionShardKey``

ajhuh-mdb: Or actually, @ajayvijayakumar123 just to double check is this supposed to be `collModrefineCollectionShardKey` or is this supposed to be `collMod` and `refineCollectionShardKey` separately?

ajhuh-mdb:
Hi @ajayvijayakumar123 👋 Could you review this docs PR for documenting the Older Version Support limitations? I also left a comment open above for you. TIA!

source/includes/fact-older-version-limitations.rst
```diff
@@ -0,0 +1,42 @@
+- Writes that produce DDL events cannot occur on the source cluster during the
```
ajayvijayakumar123, line 1: should this PR be based against `mongodb:feature-older-version-support` feature branch instead?

source/includes/fact-older-version-limitations.rst
ajayvijayakumar123: yaa you're right lol. i made a lil typo 😅

source/includes/fact-older-version-limitations.rst
```diff
@@ -0,0 +1,42 @@
...
+  CRUD events can occur during the migration.
+
+  .. note:: 
+   
+     Writes that produce DDL events on source collections outside of the 
+     :ref:`namespace filter <c2c-filtered-sync>` are allowed.
+
+- ``geoHaystack`` indexes are not supported.
+
+- :ref:`/reverse <c2c-api-reverse>` endpoints are not supported. You can't
```
ajayvijayakumar123, line 26:
```suggestion
- :ref:`/reverse <c2c-api-reverse>` endpoint is not supported. You can't 
```

ajayvijayakumar123, line 26: Are you planning to update the reverse limitation on the reverse page in a separate PR?

source/includes/fact-older-version-limitations.rst
```diff
@@ -0,0 +1,42 @@
...
+- ``geoHaystack`` indexes are not supported.
+
+- :ref:`/reverse <c2c-api-reverse>` endpoints are not supported. You can't 
+  enable the ``reversible`` option in the :ref:`/start <c2c-api-start>` request.
+
+- You can't enable the ``enableUserWriteBlocking`` option in the ``/start`` 
+  request. 
+
+- You can't enable the ``createSupportingIndexes`` :ref:`sharding parameter 
+  <c2c-api-start-sharding>` with a pre-6.0 cluster.
```
ajayvijayakumar123, line 33:
```suggestion
  <c2c-api-start-sharding>`. 
```

since this whole section is for pre-6.0 migrations

ajayvijayakumar123 (changes requested):
A couple of things!

source/includes/fact-older-version-limitations.rst
ajhuh-mdb: a h, thanks for catching that — fixed.

source/includes/fact-older-version-limitations.rst
```diff
@@ -0,0 +1,43 @@
+- Writes that produce DDL events cannot occur on the source cluster during the
```
mvankeulen94, line 1: can we add a link to the definition of DDL?

source/includes/fact-older-version-limitations.rst
```diff
@@ -0,0 +1,43 @@
...
+   
+     Writes that produce DDL events on source collections outside of the 
+     :ref:`namespace filter <c2c-filtered-sync>` are allowed.
+
+- ``geoHaystack`` indexes are not supported.
+
+- :ref:`/reverse <c2c-api-reverse>` endpoint is not supported. You can't 
+  enable the ``reversible`` option in the :ref:`/start <c2c-api-start>` request.
+
+- You can't enable the ``enableUserWriteBlocking`` option in the ``/start``
```
mvankeulen94, line 30: can we add a note here that the user is responsible for preventing writes to the source or destination?

source/includes/fact-older-version-limitations.rst
```diff
@@ -0,0 +1,43 @@
...
+- ``geoHaystack`` indexes are not supported.
+
+- :ref:`/reverse <c2c-api-reverse>` endpoint is not supported. You can't 
+  enable the ``reversible`` option in the :ref:`/start <c2c-api-start>` request.
+
+- You can't enable the ``enableUserWriteBlocking`` option in the ``/start`` 
+  request. 
+
+- You can't enable the ``createSupportingIndexes`` :ref:`sharding parameter 
+  <c2c-api-start-sharding>`.
```
mvankeulen94, line 34: @ajayvijayakumar123 is there a workaround we can advise for users? It's tricky to digest what this parameter does if the reader isn't already familiar with it

source/includes/fact-older-version-limitations.rst
```diff
@@ -0,0 +1,43 @@
...
+
+- You can't enable the ``enableUserWriteBlocking`` option in the ``/start`` 
+  request. 
+
+- You can't enable the ``createSupportingIndexes`` :ref:`sharding parameter 
+  <c2c-api-start-sharding>`. 
+
+- Indexes with inconsistent specs or that are missing on some shards are not 
+  supported. To check for index inconsistencies, see 
+  :ref:`manage-indexes-find-inconsistent-indexes`.
```
mvankeulen94, line 38: @ajayvijayakumar123 is this a change from status quo mongosync GA? We have 'mongosync only syncs indexes that have consistent index specifications on all shards.' in the limitations today

mvankeulen94 (changes requested):
thanks for adding this! a couple comments from my end

source/includes/fact-older-version-limitations.rst
ajhuh-mdb: @mvankeulen94 @ajayvijayakumar123 do we have a way of ensuring users don't write to their clusters _without_ the `enableUserWriteBlocking` option?

ajayvijayakumar123: I think they just need to manually make sure

source/includes/fact-older-version-limitations.rst
ajayvijayakumar123: Yup. Let's say the user wants to shard on shard key `a:1`. Instead of enabling `createSupportingIndexes,` the user would need to create an index to support the shard key on the source.

source/includes/fact-older-version-limitations.rst
ajayvijayakumar123: Yup. For 6.0+ migrations, mongosync _ignores_ inconsistent indexes and just doesn't build them on the destination. For pre-6.0 migrations, mongosync will error on these indexes.

ajhuh-mdb: I'll make it clear that it'll error on pre-6.0 migrations

mvankeulen94: thanks!

mvankeulen94 (approved):
thanks! lgtm

PR #533, opened 2024-12-23
DOCSP-45861 Older Version Support Task Page

ajhuh-mdb:


source/includes/table-pre-6.0-compatibilty.rst
```diff
@@ -0,0 +1,24 @@
+.. list-table::
```
jmd-mongo, line 1: [praise]

Thanks for this! This presentation is easily understood at a glance.

source/reference/older-version-support.txt
```diff
@@ -0,0 +1,93 @@
...
+
+.. include:: /includes/table-patch-versions.rst
+
+To learn more about supported server versions, see 
+:ref:`c2c-server-version-compatibility`.
+
+Steps 
+-----
+
+In the following example, ``cluster0`` is an source cluster running MongoDB 6.0,
```
jmd-mongo, line 50: [nit]

This ought to be "a source cluster" or "the source cluster".

jmd-mongo (approved):
Thanks for these updates, @ajhuh-mdb! This one LGTM mod a nit.

Thanks,
Joe

source/reference/older-version-support.txt
ajhuh-mdb: ah typo that I left in from an earlier iteration. updating!

ajhuh-mdb:
Hey @ajayvijayakumar123 👋 Whenever you get the chance, could you review this docs PR for the older version support page? TIA!

source/reference/older-version-support.txt
```diff
@@ -0,0 +1,93 @@
...
+
+Ensure that you are at least using a compatible patch version on your 
+source and destination clusters: 
+
+.. include:: /includes/table-patch-versions.rst
+
+To learn more about supported server versions, see 
+:ref:`c2c-server-version-compatibility`.
+
+Steps
```
ajayvijayakumar123, line 47: I think we should recommend using mongomirror or mongodump / mongorestore if the user can't upgrade their source cluster to 4.4+.

@mvankeulen94, you mentioned that we should recommend users to use unsupported older versions of mongodump and mongorestore to perform a skip-level upgrade on [https://jira.mongodb.org/browse/DOCSP-41415](https://www.google.com/url?q=https://jira.mongodb.org/browse/DOCSP-41415&sa=D&source=docs&ust=1736181940897321&usg=AOvVaw1INWDtEJBKa_wuP8eX7tcn) 's comments. I'm not sure if we should publicly recommend using unsupported versions of dump/restore. Thoughts?

source/reference/older-version-support.txt
```diff
@@ -0,0 +1,93 @@
...
+
+      .. literalinclude:: /includes/api/responses/success.json
+         :language: json
+         :copyable: false
+
+Learn More 
+----------
+
+- :ref:`c2c-connecting`
+- :ref:`Pre-6.0 Migration Limitations <c2c-older-version-limitations>`
```
ajayvijayakumar123, line 92: I think we should further call out the pre-6.0 migration limitations. Can we include this in the `Before You Begin` section?

source/reference/older-version-support.txt
```diff
@@ -0,0 +1,93 @@
...
+
+.. include:: /includes/table-patch-versions.rst
+
+To learn more about supported server versions, see 
+:ref:`c2c-server-version-compatibility`.
+
+Steps 
+-----
+
+In the following example, ``cluster0`` is a source cluster running MongoDB 6.0,
```
ajayvijayakumar123, line 50:
```suggestion
In the following example, ``cluster0`` is a source cluster running MongoDB 5.0, 
```

Since this page talks about pre-6.0 migrations

source/reference/older-version-support.txt
ajhuh-mdb: Ah got it. Similar to what you mentioned, I wasn't sure if we wanted to recommend that in our docs since it seemed like more of a workaround rather than an official solution. Can wait for Maria's comments on this!

source/reference/older-version-support.txt
ajhuh-mdb: Great catch, fixing.

source/reference/older-version-support.txt
ajhuh-mdb: Will do!

source/reference/older-version-support.txt
mvankeulen94: Thanks all, agreed that we shouldn't publicly document using unsupported versions of dump+restore. In the meantime since my July comment, we have introduced some internal guidance around migration from EOL'd versions. We can leave the public guidance scoped to officially supported tooling

source/reference/older-version-support.txt
```diff
@@ -0,0 +1,98 @@
...
+
+Before You Begin 
+----------------
+
+:ref:`Download and install <c2c-install>` ``mongosync`` 1.10 or later.
+
+Embedded Verifier 
+~~~~~~~~~~~~~~~~~
+
+When you sync from a pre-6.0 source cluster, enable the
```
mvankeulen94, line 33: @ajayvijayakumar123 is this strictly necessary? Now that embedded verification is decoupled from the persistence project, there are still migrations that wouldn't be able to proceed with verification. Can we remove this embedded verifier section and just link to the verify data transfer docs instead?

source/reference/older-version-support.txt
ajayvijayakumar123: @mvankeulen94 to confirm, we don't want to mention dump/restore or mongomirror on this page?

source/reference/older-version-support.txt
ajayvijayakumar123: Yup that works. I just wanted us to recommend users to verify pre-6.0 migrations. Linking to the verify data transfer [docs](https://www.mongodb.com/docs/cluster-to-cluster-sync/current/reference/verification/) sounds good

source/reference/older-version-support.txt
mvankeulen94: yeah we can exclude those tools from this page. thanks!

ajhuh-mdb:
Discussed with @ajayvijayakumar123 on Slack that the Older Version Support Task page may be redundant, so we'll keep that information contained in the release notes and the existing Version Compatibility and Limitations pages.

ajayvijayakumar123 (approved):
LGTM!

PR #557, opened 2025-01-21
DOCSP-45863-embedded-verifier-migrations

ajhuh-mdb:
## DESCRIPTION
Adds a note that during pre-6.0 migrations, the verifier checks metadata and indexes _only_ for DDL events. 
- Opted to keep the previously documented limitations in their own section, remove the redundant note at the top of the page, and make the new functionality a note since it doesn't apply to all migrations.

## STAGING 
https://deploy-preview-557--docs-cluster-to-cluster-sync.netlify.app/reference/verification/embedded/#unsupported-verification-checks

## JIRA
https://jira.mongodb.org/browse/DOCSP-45863

### DIFF
--- a/source/includes/fact-verifier-limitations.rst
+++ b/source/includes/fact-verifier-limitations.rst
@@ -32,11 +32,6 @@ The embedded verifier has the following limitations:
 
 - .. include:: /includes/fact-verifier-buildIndexes
 
-.. note:: 
-
-   Starting in version 1.10, the verifier checks metadata and indexes 
-   for DDL events only when migrating from a pre-6.0 source cluster.
-
 Unsupported Verification Checks
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
--- a/source/includes/fact-verifier-limitations.rst
+++ b/source/includes/fact-verifier-limitations.rst
@@ -41,10 +41,11 @@ Unsupported Verification Checks
 
    Starting in version 1.10, the verifier checks for data inconsistencies from 
    a :ref:`DDL event <c2c-older-version-limitations>` that occurred on the 
-   pre-6.0 source cluster during migration.
+   pre-6.0 source cluster during migration. This is because pre-6.0 migrations 
+   do not support DDL events.
 
-   To learn more about pre-6.0 migrations and DDL events, see 
-   :ref:`Pre-6.0 Migration Limitations <c2c-older-version-limitations>`.
+   To learn more, see :ref:`Pre-6.0 Migration Limitations 
+   <c2c-older-version-limitations>`.
 
 Unsupported Beta Features
 ~~~~~~~~~~~~~~~~~~~~~~~~~
--- a/source/includes/fact-verifier-limitations.rst
+++ b/source/includes/fact-verifier-limitations.rst
@@ -39,8 +39,12 @@ Unsupported Verification Checks
 
 .. note:: 
 
-   Starting in version 1.10, the verifier checks metadata and indexes 
-   for DDL events only when migrating from a pre-6.0 source cluster.
+   Starting in version 1.10, the verifier checks for data inconsistencies from 
+   a :ref:`DDL event <c2c-older-version-limitations>` that occurred on the 
+   pre-6.0 source cluster during migration.
+
+   To learn more about pre-6.0 migrations and DDL events, see 
+   :ref:`Pre-6.0 Migration Limitations <c2c-older-version-limitations>`.
 
 Unsupported Beta Features
 ~~~~~~~~~~~~~~~~~~~~~~~~~
--- a/source/includes/fact-verifier-limitations.rst
+++ b/source/includes/fact-verifier-limitations.rst
@@ -32,10 +32,20 @@ The embedded verifier has the following limitations:
 
 - .. include:: /includes/fact-verifier-buildIndexes
 
+.. note:: 
+
+   Starting in version 1.10, the verifier checks metadata and indexes 
+   for DDL events only when migrating from a pre-6.0 source cluster.
+
 Unsupported Verification Checks
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
 .. include:: /includes/fact-verifier-unsupported
+
+.. note:: 
+
+   Starting in version 1.10, the verifier checks metadata and indexes 
+   for DDL events only when migrating from a pre-6.0 source cluster.
 
 Unsupported Beta Features
 ~~~~~~~~~~~~~~~~~~~~~~~~~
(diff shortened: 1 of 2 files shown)


mdb-ashley (approved):
This LGTM, thanks Ali!

ajhuh-mdb:
Hey @ajayvijayakumar123 and @mvankeulen 👋 Whenever you get the chance, could you both review this docs PR for noting verifier changes for pre-6.0 migrations? TIA!

source/includes/fact-verifier-limitations.rst
```diff
@@ -37,6 +37,11 @@ Unsupported Verification Checks
 
 .. include:: /includes/fact-verifier-unsupported
 
+.. note:: 
+
+   Starting in version 1.10, the verifier checks metadata and indexes
```
ajayvijayakumar123, line 42: I think it would be clearer to instead say `The verifier checks if there’s a data inconsistency from a DDL event that occurred on the source when migrating from a pre-6.0 source cluster.`

 We could then link to the new pre-6.0 migrations limitations [section](https://deploy-preview-522--docs-cluster-to-cluster-sync.netlify.app/reference/limitations/#pre-6.0-migrations) or just add the bullet about DDL events to this page too.

ajayvijayakumar123: I don't think we need to mention metadata or indexes since that's an implementation detail in how we're checking for DDL events. @mvankeulen94 thoughts here?

source/includes/fact-verifier-limitations.rst
```diff
@@ -37,6 +37,11 @@ Unsupported Verification Checks
 
 .. include:: /includes/fact-verifier-unsupported
 
+.. note::
```
ajayvijayakumar123, line 40: can we also include this note:

`
“Mongosync preserves document field-order based on what it reads using read preference primary from the source cluster. The embedded verifier will verify documents based on the source cluster’s primary node. It will not be able to verify discrepancies among the different source cluster nodes. In rare cases, discrepancies among the different source cluster’s nodes will cause the embedded verifier to fail a migration.”
`

at the top of the page? This applies to _all_ migrations

ajayvijayakumar123 (changes requested):
Some suggestions!

source/includes/fact-verifier-limitations.rst
rohshar: lgtm % changing `In rare cases, discrepancies among the different source cluster’s nodes will cause the embedded verifier to fail a migration` to `In rare cases, discrepancies in document field order among the source cluster’s nodes will cause the embedded verifier to fail a migration even if mongosync copied documents correctly`

It may also make sense to mention that the only case where this can happen is if an election occurs.

source/includes/fact-verifier-limitations.rst
ajhuh-mdb: gotcha! updating

ajhuh-mdb:
Thanks for the review @ajayvijayakumar123. Just finished adding in the changes you suggested. For the note that applies to all migrations, I tweaked the wording a bit, but let me know if it's changed the meaning too much. Thanks again!

source/includes/fact-verifier-limitations.rst
```diff
@@ -37,6 +37,15 @@ Unsupported Verification Checks
 
 .. include:: /includes/fact-verifier-unsupported
 
+.. note:: 
+
+   Starting in version 1.10, the verifier checks for data inconsistencies from 
+   a :ref:`DDL event <c2c-older-version-limitations>` that occurred on the
```
ajayvijayakumar123, line 43:
```suggestion
    :ref:`DDL events <c2c-older-version-limitations>` that occurred on the 
```

ajayvijayakumar123: Could we also note here that DDL events are unsupported in pre-6.0 migrations?

source/includes/fact-verifier-limitations.rst
ajayvijayakumar123: can we note that this issue can only occur in rare cases?

ajayvijayakumar123 (changes requested):
a couple more suggestions!

source/includes/fact-verifier-limitations.rst
ajhuh-mdb: Just to double check my understanding is it that _any_ time an election occurs this can happen or just a small portion of elections? @ajayvijayakumar123 @rohshar

ajayvijayakumar123: Discrepancies among different source cluster nodes only happen in very rare cases. Even if there is a discrepancy, an election may not even surface the issue and cause the verifier to fail. I think we can just say that this issue is rare in general.

ajayvijayakumar123 (approved):
LGTM! @rohshar Can you be the second reviewer for this PR? (since Maria's out sick)

source/reference/verification/embedded.txt
```diff
@@ -22,9 +22,16 @@ Verify with Embedded Verifier
...
+.. note:: 
+
+   ``mongosync`` reads using :readmode:`primary` read preference, so it 
+   preserves document field order from the source cluster's primary node. The 
+   embedded verifier also checks documents based on the source cluster’s primary 
+   node. Because of this read preference, the verifier cannot verify 
+   discrepancies between the different source cluster nodes. In rare cases, 
+   discrepancies in document field order between the source cluster’s nodes will 
+   cause the embedded verifier to fail the migration, even if ``mongosync`` 
+   copied the documents correctly.
```
rohshar, line 34:
```suggestion
   ``mongosync`` reads using :readmode:`primary` read preference, so it 
   preserves document field order from the source cluster's primary node. The 
   embedded verifier also checks documents based on the source cluster’s primary 
   node, but at a different time than mongosync reads them. Because of this, in rare cases, 
   discrepancies in document field order between the source cluster’s nodes will 
   cause the embedded verifier to fail the migration, even if ``mongosync`` 
   copied the documents correctly.
```

rohshar: what do you think about this change @ajayvijayakumar123

rohshar (changes requested):

source/reference/verification/embedded.txt
ajayvijayakumar123: yup that sounds good. should we also note that discrepancies among cluster nodes themselves are rare?

rohshar: I think that's already implied in the suggested text

ajhuh-mdb: Got it, thanks both. Updated.

rohshar (approved):

PR #591, opened 2025-02-05
(DOCSP-46983) Removes sentence in limitations for pre 6 migrations

morales-mario-mdb:


jeff-allen-mongo (approved):
This update LGTM from a copy perspective, but I wanted to call out that on the [/commit endpoint page](https://deploy-preview-591--docs-cluster-to-cluster-sync.netlify.app/reference/api/commit/#requirements) we still have a bullet under requirements that states:

> Stop your application to prevent any further writes to the source cluster. During commit, you may still read from the source cluster.
> 
> Warning
> 
> If you write to the source cluster during commit, you might experience data loss.

I'm wondering if we need to remove that bullet as well? It would be good to get product / engineering input on this.

morales-mario-mdb:
The PR appears to no longer be required. Asked Jane Tio to review and she replied:
> Hey Mario! I actually don't think we need the PR anymore, since the docs were changed from when I created the ticket. The sentence about being able to write to the source post /commit is accurate now.
Should we close the ticket? @jeff-allen-mongo

morales-mario-mdb:
@jeff-allen-mongo  Just confirmed with Jane, this PR is no longer required. The docs have changed to: 

> Ensure that no writes are made to the source cluster after you call the /commit endpoint.

The sentence that was originally wrong was:

> Ensure that no writes are made to the source or destination cluster during the migration.

PR #599, opened 2025-02-06
DOCSP-47159: Clarify write blocking recommendation on finalize cutover page

mayaraman19:
https://jira.mongodb.org/browse/DOCSP-47159

Staging:

https://deploy-preview-599--docs-cluster-to-cluster-sync.netlify.app/reference/cutover-process/#if-you-manually-blocked-writes-on-the-destination-cluster-by-using---enable-application-writes-on-the-destination-cluster.

### DIFF
--- a/source/reference/cutover-process.txt
+++ b/source/reference/cutover-process.txt
@@ -179,7 +179,7 @@ Steps
 
       Then, transfer your application workload to the destination cluster.
 
-      If you start ``mongosync``
+      If you started ``mongosync``
       with write-blocking by using the ``enableUserWriteBlocking`` option on
       the :ref:`/start <c2c-api-start>` endpoint, you do not need
       to complete this step.
--- a/source/reference/cutover-process.txt
+++ b/source/reference/cutover-process.txt
@@ -164,7 +164,9 @@ Steps
 
       For more information, see :ref:`c2c-verification`.
 
-   .. step:: Enable application writes on the destination cluster.
+   .. step:: If you did not start ``mongosync`` with 
+      :ref:`write-blocking <c2c-write-blocking>`, 
+      enable application writes on the destination cluster.
 
       To enable writes, update :dbcommand:`setUserWriteBlockMode`:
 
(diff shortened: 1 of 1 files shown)


gmiller-mdb (approved):
LGTM!

source/reference/cutover-process.txt
```diff
@@ -164,7 +164,7 @@ Steps
 
       For more information, see :ref:`c2c-verification`.
 
-   .. step:: Enable application writes on the destination cluster.
+   .. step:: If you did not start ``mongosync`` with :ref:`write-blocking <c2c-write-blocking>`, enable application writes on the destination cluster.
```
ajayvijayakumar123, line 167: The user also only needs to do this step if they manually blocked writes on the dst. Could we update the step title to be something like "if you blocked writes on the destination cluster, enable application writes?" We can still keep the other note you added below

source/reference/cutover-process.txt
```diff
@@ -179,6 +179,11 @@ Steps
 
       Then, transfer your application workload to the destination cluster.
 
+      If you start ``mongosync``
```
ajayvijayakumar123, line 182:
```suggestion
      If you started ``mongosync``
```

ajayvijayakumar123 (changes requested):
Some suggestions!

source/reference/cutover-process.txt
```diff
@@ -164,7 +164,7 @@ Steps
 
       For more information, see :ref:`c2c-verification`.
 
-   .. step:: Enable application writes on the destination cluster.
+   .. step:: If you manually :ref:`blocked writes <c2c-write-blocking>` on the destination cluster by using :dbcommand:`setUserWriteBlockMode`, enable application writes on the destination cluster.
```
ajayvijayakumar123, line 167:
```suggestion
   .. step:: If you manually blocked writes on the destination cluster by using :dbcommand:`setUserWriteBlockMode`, enable application writes on the destination cluster.
```

The link could be confusing because the text references manually blocking writes but the link references mongosync write-blocking (which is different from the manual option)

ajayvijayakumar123 (changes requested):
1 small suggestion. can LGTM right after!

ajayvijayakumar123 (approved):
LGTM!

PR #670, opened 2025-03-14
DOCSP-46644: Stop balancer on a per-collection basis

mayaraman19:
https://jira.mongodb.org/browse/DOCSP-46644

Staging:
- https://deploy-preview-670--docs-cluster-to-cluster-sync.netlify.app/reference/mongosync/mongosync-behavior/#sharded-clusters
- https://deploy-preview-670--docs-cluster-to-cluster-sync.netlify.app/reference/limitations/#sharded-clusters

### DIFF
--- a/source/includes/fact-mongosync-balancer.rst
+++ b/source/includes/fact-mongosync-balancer.rst
@@ -12,4 +12,5 @@
    filter, do not run :dbcommand:`shardCollection` on collections
    within the namespace filter. If you run :dbcommand:`shardCollection` on 
    collections within the namespace filter during the migration, ``mongosync``
-   returns an error and stops.
+   returns an error and stops, which requires you to start the migration
+   from scratch.
--- a/source/includes/fact-mongosync-balancer.rst
+++ b/source/includes/fact-mongosync-balancer.rst
@@ -1,21 +1,21 @@
 .. important::
 
    If the source or destination cluster is a sharded cluster
-   and you are not running ``mongosync`` with namespace filtering,
+   and you are not running ``mongosync`` with :ref:`namespace
+   filtering <c2c-filtered-sync>`,
    you must disable the source cluster's balancer
    by running the :dbcommand:`balancerStop` command and waiting 15 minutes
    for the command to complete.
 
    If the source or destination cluster is a sharded cluster and you
-   are running ``mongosync`` with :ref:`namespace
-   filtering <c2c-filtered-sync>`, you can enable the source cluster's
+   are running ``mongosync`` with namespace filtering, you can enable the source cluster's
    balancer overall but you must disable it for 
    all collections within the namespace filter. 
    See :ref:`disabling-balancer-filtered`. You can also disable
    the source cluster's balancer.
 
    You must always disable the balancer on a sharded destination
-   cluster.
+   cluster by using :dbcommand:`balancerStop`.
 
    During migration, do not run the :dbcommand:`moveChunk` or 
    :dbcommand:`moveRange` commands. If you have enabled the source cluster's
--- a/source/includes/fact-mongosync-balancer.rst
+++ b/source/includes/fact-mongosync-balancer.rst
@@ -1,10 +1,21 @@
 .. important::
 
-   When the source or destination cluster is a sharded cluster, you must stop 
-   the source cluster's balancer on collections within the :ref:`namespace
-   filter <c2c-filtered-sync>`. You can enable the source cluster's balancer on
-   the cluster level and for collections outside the namespace filter. 
-   You must also stop the destination cluster's balancer.
+   If the source or destination cluster is a sharded cluster
+   and you are not running ``mongosync`` with namespace filtering,
+   you must disable the source cluster's balancer
+   by running the :dbcommand:`balancerStop` command and waiting 15 minutes
+   for the command to complete.
+
+   If the source or destination cluster is a sharded cluster and you
+   are running ``mongosync`` with :ref:`namespace
+   filtering <c2c-filtered-sync>`, you can enable the source cluster's
+   balancer overall but you must disable it for 
+   all collections within the namespace filter. 
+   See :ref:`disabling-balancer-filtered`. You can also disable
+   the source cluster's balancer.
+
+   You must always disable the balancer on a sharded destination
+   cluster.
 
    During migration, do not run the :dbcommand:`moveChunk` or 
    :dbcommand:`moveRange` commands. If you have enabled the source cluster's
--- a/source/includes/fact-mongosync-balancer.rst
+++ b/source/includes/fact-mongosync-balancer.rst
@@ -1,7 +1,24 @@
 .. important::
 
    When the source or destination cluster is a sharded cluster, you must stop 
-   the balancer on both clusters and not run the :dbcommand:`moveChunk` or
-   :dbcommand:`moveRange` commands for the duration of the migration. To stop 
-   the balancer, run the :dbcommand:`balancerStop` command and wait for the 
-   command to complete.
+   the source cluster's balancer on collections within the :ref:`namespace
+   filter <c2c-filtered-sync>`. You can enable the source cluster's balancer on
+   the cluster level and for collections outside the namespace filter. 
+   You must also stop the destination cluster's balancer.
+
+   If you enable the source cluster's balancer but do not use a
+   namespace filter, or if you disable the balancer for all
+   collections within the namespace filter, ``mongosync`` fails.
+
+   During migration, do not run the :dbcommand:`moveChunk` or 
+   :dbcommand:`moveRange` commands. If you have enabled the source cluster's
+   balancer, but disabled it for collections within the namespace
+   filter, do not run :dbcommand:`shardCollection` on collections
+   within the namespace filter.
+
+   To stop the balancer, run the :dbcommand:`balancerStop` command 
+   and wait for the command to complete.
+     
+   After stopping the balancer, wait fifteen minutes before
+   starting ``mongosync``. This gives the cluster time to
+   finish any in progress chunk migrations.
--- a/source/includes/fact-mongosync-balancer.rst
+++ b/source/includes/fact-mongosync-balancer.rst
@@ -6,19 +6,10 @@
    the cluster level and for collections outside the namespace filter. 
    You must also stop the destination cluster's balancer.
 
-   If you enable the source cluster's balancer but do not use a
-   namespace filter, or if you disable the balancer for all
-   collections within the namespace filter, ``mongosync`` fails.
-
    During migration, do not run the :dbcommand:`moveChunk` or 
    :dbcommand:`moveRange` commands. If you have enabled the source cluster's
    balancer, but disabled it for collections within the namespace
    filter, do not run :dbcommand:`shardCollection` on collections
-   within the namespace filter.
-
-   To stop the balancer, run the :dbcommand:`balancerStop` command 
-   and wait for the command to complete.
-     
-   After stopping the balancer, wait fifteen minutes before
-   starting ``mongosync``. This gives the cluster time to
-   finish any in progress chunk migrations.
+   within the namespace filter. If you run :dbcommand:`shardCollection` on 
+   collections within the namespace filter during the migration, ``mongosync``
+   returns an error and stops.
--- a/source/includes/fact-mongosync-balancer.rst
+++ b/source/includes/fact-mongosync-balancer.rst
@@ -1,5 +1,11 @@
 .. important::
 
+   You must always disable the balancer on a sharded destination
+   cluster by using :dbcommand:`balancerStop`. 
+   After stopping the balancer, wait **fifteen minutes** before
+   starting ``mongosync``. This gives the cluster time to
+   finish any in-progress chunk migrations.
+
    If the source or destination cluster is a sharded cluster
    and you are not running ``mongosync`` with :ref:`namespace
    filtering <c2c-filtered-sync>`,
@@ -8,14 +14,12 @@
    for the command to complete.
 
    If the source or destination cluster is a sharded cluster and you
-   are running ``mongosync`` with namespace filtering, you can enable the source cluster's
-   balancer overall but you must disable it for 
+   are running ``mongosync`` with namespace filtering, you can 
+   globally enable the source cluster's
+   balancer but you must disable it for 
    all collections within the namespace filter. 
-   See :ref:`disabling-balancer-filtered`. You can also disable
+   See :ref:`disabling-balancer-filtered`. You can also fully disable
    the source cluster's balancer.
-
-   You must always disable the balancer on a sharded destination
-   cluster by using :dbcommand:`balancerStop`.
 
    During migration, do not run the :dbcommand:`moveChunk` or 
    :dbcommand:`moveRange` commands. If you have enabled the source cluster's
--- a/source/reference/limitations.txt
+++ b/source/reference/limitations.txt
@@ -125,21 +125,21 @@ Sharded Clusters
      Indexes Across Shards <manage-indexes-find-inconsistent-indexes>`.
 
 - If the source or destination cluster is a sharded cluster
-  and you are not running ``mongosync`` with namespace filtering,
+  and you are not running ``mongosync`` with :ref:`namespace
+  filtering <c2c-filtered-sync>`,
   you must disable the source cluster's balancer
   by running the :dbcommand:`balancerStop` command and waiting 15 minutes
   for the command to complete.
 
   If the source or destination cluster is a sharded cluster and you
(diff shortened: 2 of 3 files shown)


source/includes/fact-mongosync-balancer.rst
```diff
@@ -1,7 +1,24 @@
...
+   balancer, but disabled it for collections within the namespace
+   filter, do not run :dbcommand:`shardCollection` on collections
+   within the namespace filter.
+
+   To stop the balancer, run the :dbcommand:`balancerStop` command 
+   and wait for the command to complete.
+     
+   After stopping the balancer, wait fifteen minutes before
+   starting ``mongosync``. This gives the cluster time to
+   finish any in progress chunk migrations.
```
mayaraman19, line 24: int reviewer: would love feedback on whether this would be better if taken out of the `.. important::`

source/reference/limitations.txt
```diff
@@ -124,16 +124,20 @@ Sharded Clusters
...
+  If you are not running a filtered sync, you must stop the source cluster's
+  balancer.
+  You must also stop the destination cluster's balancer. See 
+  :ref:`enabling-balancer-filtered` to learn how to enable the balancer during
+  filtered sync.
+
+- If you have enabled the source cluster's
+  balancer, but disabled it for collections within the namespace
+  filter, do not run :dbcommand:`shardCollection` on collections
+  within the namespace filter.
```
gmiller-mdb, line 140: Is there a description of what will go wrong/error message for this? It's very clear, but I am just thinking about the scenario where a user does run this command and then googles their error.

source/includes/fact-mongosync-balancer.rst
gmiller-mdb: I am wondering if the second and third paragraphs can be moved out of the important to below it? or simply move the final two paragraphs up (putting all of the information about stopping the balancer together) and then seeing how that reads.

gmiller-mdb (commented):
mostly LTGM, with a few questions!

source/reference/limitations.txt
mayaraman19: @gmiller-mdb I think we tend not to write error messages into the docs, but I can describe a bit more about what happens?

gmiller-mdb: @mayaraman19 that sounds great to me! I think one short sentence should cover it.

gmiller-mdb (commented):
one comment

gmiller-mdb (approved):
LGTM!

source/includes/fact-mongosync-balancer.rst
```diff
@@ -1,7 +1,16 @@
 .. important::
 
    When the source or destination cluster is a sharded cluster, you must stop 
-   the balancer on both clusters and not run the :dbcommand:`moveChunk` or
-   :dbcommand:`moveRange` commands for the duration of the migration. To stop 
-   the balancer, run the :dbcommand:`balancerStop` command and wait for the 
-   command to complete.
+   the source cluster's balancer on collections within the :ref:`namespace
```
nickweinberger, line 4: I think it might help to be a bit more clear about the distinction between the namespace filter case and the non-namespace filter case, since per-collection balancing is only an option when mongosync is run with the namespace filter. 

Maybe something like "When running mongosync with namespace filtering and the source or destination cluster is a sharded cluster, you must disable the source cluster's balancer for all collections within the namespace filter."

And then adding a note for the non-namespace filtering, eg: "If running mongosync without namespace filtering, you must stop the balancer on both clusters."

Does this make sense?

source/reference/limitations.txt
```diff
@@ -124,16 +124,20 @@ Sharded Clusters
...
-  for the entire lifetime of a migration. To stop the balancer, run the
-  :dbcommand:`balancerStop` command and wait for the command to
-  complete.
-
-  .. note::
-
-     After stopping the balancer, wait fifteen minutes before
-     starting ``mongosync``. This gives the cluster time to
-     finish any in progress chunk migrations.
+- When the source or destination cluster is a sharded cluster, you must stop
```
nickweinberger, line 127: Similar to above–I think the "If you are not running a filtered sync, you must stop the source cluster's balancer" helps here, but I would still suggest adding a note upfront saying "if running mognosync with a namespace filter..." just to make it clear that there are two cases.
a) with ns filter -> per collection balancing is an option on the source
b) wihtout ns filter -> both balancer must be off

Also, would it be possible to update the language just to make the following clear: even if there's a ns filter, the user doesn't have to do per collection balancing. This is just an option–they _can_ leave the source balancer enabled for collections outside the ns filter, but also, if they disable the source balancer completely that works too.

source/reference/mongosync/mongosync-behavior.txt
```diff
@@ -102,6 +102,40 @@ preserve the source cluster's sharding configuration.
...
+
+Then, run the :dbcommand:`balancerStop` command to stop
+the balancer on the destination cluster and wait for the command to complete.
+
+After stopping the balancer, wait fifteen minutes before
+starting ``mongosync``. This gives the cluster time to
+finish any in progress chunk migrations.
+
+If you enable the source cluster's balancer but do not use a
+namespace filter, or if you disable the balancer for all
```
nickweinberger, line 136: Think you're missing a "not" here:

"...if you do **not** disable the balancer for all collections within the namespace filter, mongosync fails".

source/reference/mongosync/mongosync-behavior.txt
```diff
@@ -102,6 +102,40 @@ preserve the source cluster's sharding configuration.
...
+     }
+   )
+
+Run the preceding command for every collection within the
+namespace filter.
+
+Then, run the :dbcommand:`balancerStop` command to stop
+the balancer on the destination cluster and wait for the command to complete.
+
+After stopping the balancer, wait fifteen minutes before
```
nickweinberger, line 131: Would suggest adding this to the end of the last paragraph, but up to you!

source/reference/mongosync/mongosync-behavior.txt
```diff
@@ -102,6 +102,40 @@ preserve the source cluster's sharding configuration.
 
 .. include:: /includes/fact-mongosync-balancer.rst
 
+.. _enabling-balancer-filtered:
+
+Enabling Balancer in Filtered Sync
+''''''''''''''''''''''''''''''''''
+
+Before starting ``mongosync`` with a :ref:`namespace filter
```
nickweinberger, line 110: Here i think it's important to clarify that this is optional. The message being "if you want to leave the balancer enabled for collections outside the namespace filter, here's how you can ensure the balancer is disabled for collections within the namespace filter, which is mongosync's requirement."

To that end, I think it might make more sense to label the title above "Disabling the balancer for collections for filtered sync" rather than "Enabling Balancer in Filtered Sync"

nickweinberger (changes requested):
I left a few suggestions–please let me know if you have any questions or if I can clarify anything!

source/reference/limitations.txt
```diff
@@ -124,16 +124,30 @@ Sharded Clusters
...
+- If the source or destination cluster is a sharded cluster
+  and you are not running ``mongosync`` with :ref:`namespace
+  filtering <c2c-filtered-sync>`,
+  you must disable the source cluster's balancer
+  by running the :dbcommand:`balancerStop` command and waiting 15 minutes
+  for the command to complete.
+
+  If the source or destination cluster is a sharded cluster and you
+  are running ``mongosync`` with namespace filtering, you can enable the source cluster's
+  balancer overall but you must disable it for
```
nickweinberger, line 136: nit: I slightly prefer "globally" or "at the cluster level" to "overall", but this isn't blocking. Up to you.

source/includes/fact-mongosync-balancer.rst
```diff
@@ -1,7 +1,27 @@
...
+   If the source or destination cluster is a sharded cluster
+   and you are not running ``mongosync`` with :ref:`namespace
+   filtering <c2c-filtered-sync>`,
+   you must disable the source cluster's balancer
+   by running the :dbcommand:`balancerStop` command and waiting 15 minutes
+   for the command to complete.
+
+   If the source or destination cluster is a sharded cluster and you
+   are running ``mongosync`` with namespace filtering, you can enable the source cluster's
+   balancer overall but you must disable it for
```
nickweinberger, line 12: same as other comment on "overall"

source/includes/fact-mongosync-balancer.rst
```diff
@@ -1,7 +1,27 @@
...
+   filtering <c2c-filtered-sync>`,
+   you must disable the source cluster's balancer
+   by running the :dbcommand:`balancerStop` command and waiting 15 minutes
+   for the command to complete.
+
+   If the source or destination cluster is a sharded cluster and you
+   are running ``mongosync`` with namespace filtering, you can enable the source cluster's
+   balancer overall but you must disable it for 
+   all collections within the namespace filter. 
+   See :ref:`disabling-balancer-filtered`. You can also disable
```
nickweinberger, line 14:
```suggestion
   See :ref:`disabling-balancer-filtered`. You can also fully disable
```

source/reference/mongosync/mongosync-behavior.txt
```diff
@@ -102,6 +102,59 @@ preserve the source cluster's sharding configuration.
 
 .. include:: /includes/fact-mongosync-balancer.rst
 
+.. _disabling-balancer-filtered:
+
+Disabling Balancer for Collections in Filtered Sync
+'''''''''''''''''''''''''''''''''''''''''''''''''''
+
+If you are using a :ref:`namespace filter <c2c-filtered-sync>`
+and want to enable your source cluster's balancer,
```
nickweinberger, line 111:
```suggestion
and want to enable your source cluster's balancer for collections outside the namespace filter, 
```

Small suggestion just to align this with what would motivate a customer to do this–it's all about leaving the balancer enabled for those non-migrated collections (those outside the ns filter)!

source/reference/mongosync/mongosync-behavior.txt
```diff
@@ -102,6 +102,59 @@ preserve the source cluster's sharding configuration.
...
+use the following instructions before
+starting ``mongosync``. These instructions allow you to enable the balancer 
+for the source cluster and disable the balancer for collections 
+within the namespace filter.
```
nickweinberger, line 115:
```suggestion
starting ``mongosync``.
```
Up to you but I'm thinking it may actually be more clear without this last sentence.

source/reference/mongosync/mongosync-behavior.txt
```diff
@@ -102,6 +102,59 @@ preserve the source cluster's sharding configuration.
...
+          { 
+            setAllowMigrations: “<db>.<collection>”,
+            allowMigrations: false
+          }
+        )
+
+      Run the preceding command for every collection within the
+      namespace filter.
+
+   .. step:: Stop the balancer on the destination cluster.
```
nickweinberger, line 143: I'm not sure this is needed here since it already says at the top under **Important**:  "You must always disable the balancer on a sharded destination cluster by using [balancerStop.](https://www.mongodb.com/docs/master/reference/command/balancerStop/#mongodb-dbcommand-dbcmd.balancerStop)" 

I think it may be a bit confusing to put the mandatory instructions for disabling the destination balancer under this section specifically related to [Disabling Balancer for Collections in Filtered Sync](https://deploy-preview-670--docs-cluster-to-cluster-sync.netlify.app/reference/mongosync/mongosync-behavior/#disabling-balancer-for-collections-in-filtered-sync). Instead I think it might be easier to read if we added the info about _waiting 15 minutes before starting mongosync so chunk migrations can finish_ to the section above where we already mention disabling the balancer on a sharded destination cluster, and remove this section. 

It may even make the most sense to put this whole section starting with "You must always disable the balancer on a sharded destination cluster by using [balancerStop.](https://www.mongodb.com/docs/master/reference/command/balancerStop/#mongodb-dbcommand-dbcmd.balancerStop)" and then including the waiting for 15 minutes instructions at the **very top** (above the instructions for the source balancer) just so it's extra clear that the destination cluster's balancer must _always_ be disabled, but that users have options for the source cluster balancer.

nickweinberger (changes requested):
Left a few more comments for you, Maya. Mostly small suggestions to make the wording as clear as possible, and one larger structural suggestion to clarify the requirements for the destination cluster's balancer, so that piece doesn't get lost. Thanks for the back and forth on this!

source/includes/fact-mongosync-balancer.rst
```diff
@@ -1,7 +1,31 @@
...
-   command to complete.
+   You must always disable the balancer on a sharded destination
+   cluster by using :dbcommand:`balancerStop`. 
+   After stopping the balancer, wait **fifteen minutes** before
+   starting ``mongosync``. This gives the cluster time to
+   finish any in-progress chunk migrations.
+
+   If the source or destination cluster is a sharded cluster
+   and you are not running ``mongosync`` with :ref:`namespace
+   filtering <c2c-filtered-sync>`,
```
nickweinberger, line 11:
```suggestion
   and you are running ``mongosync`` without :ref:`namespace
   filtering <c2c-filtered-sync>`,
```

source/includes/fact-mongosync-balancer.rst
```diff
@@ -1,7 +1,31 @@
...
+   by running the :dbcommand:`balancerStop` command and waiting 15 minutes
+   for the command to complete.
+
+   If the source or destination cluster is a sharded cluster and you
+   are running ``mongosync`` with namespace filtering, you can 
+   globally enable the source cluster's
+   balancer but you must disable it for 
+   all collections within the namespace filter. 
+   See :ref:`disabling-balancer-filtered`. You can also fully disable
+   the source cluster's balancer.
```
nickweinberger, line 22: Perfect! thanks for adding this.

source/includes/fact-mongosync-balancer.rst
```diff
@@ -1,7 +1,31 @@
...
+
+   During migration, do not run the :dbcommand:`moveChunk` or 
+   :dbcommand:`moveRange` commands. If you have enabled the source cluster's
+   balancer, but disabled it for collections within the namespace
+   filter, do not run :dbcommand:`shardCollection` on collections
+   within the namespace filter. If you run :dbcommand:`shardCollection` on 
+   collections within the namespace filter during the migration, ``mongosync``
```
nickweinberger, line 29:
```suggestion
   filter, running :dbcommand:`shardCollection` on collections within the namespace filter during the migration causes ``mongosync`` to return
```

If possible to make these two sentences into one I think it might be cleaner, but up to your judgement–if you like to be explicit "do not run X command" that's ok too.

source/reference/mongosync/mongosync-behavior.txt
```diff
@@ -102,6 +102,49 @@ preserve the source cluster's sharding configuration.
...
 
+.. _disabling-balancer-filtered:
+
+Disabling Balancer for Collections in Filtered Sync
+'''''''''''''''''''''''''''''''''''''''''''''''''''
+
+If you are using a :ref:`namespace filter <c2c-filtered-sync>`
+and want to enable your source cluster's balancer for 
+collections outside the namespace filter,
+use the following instructions before
```
nickweinberger, line 113:
```suggestion
follow these instructions before
```

source/reference/limitations.txt
```diff
@@ -124,16 +124,31 @@ Sharded Clusters
...
-  :dbcommand:`balancerStop` command and wait for the command to
-  complete.
-
-  .. note::
-
-     After stopping the balancer, wait fifteen minutes before
-     starting ``mongosync``. This gives the cluster time to
-     finish any in progress chunk migrations.
+- If the source or destination cluster is a sharded cluster
+  and you are not running ``mongosync`` with :ref:`namespace
```
nickweinberger, line 128: same as other comment: I'd suggest "you are running mongosync without..."

source/reference/limitations.txt
```diff
@@ -124,16 +124,31 @@ Sharded Clusters
...
+  See :ref:`disabling-balancer-filtered`. You can also fully disable
+  the source cluster's balancer.
+
+  You must always disable the balancer on a sharded destination
+  cluster by using :dbcommand:`balancerStop`.
+
+- If you have enabled the source cluster's
+  balancer, but disabled it for collections within the namespace
+  filter, do not run :dbcommand:`shardCollection` on collections
+  within the namespace filter. If you run :dbcommand:`shardCollection` on
```
nickweinberger, line 148: Same as other comment on Behavior page–if possible to combine into one sentence to reduce repetition that would be great, but up to your discretion.

nickweinberger (approved):
✅  LGTM! I left a few non-blocking comments and suggestions just to make it as clean and clear as possible, but I'll leave those up to your judgement. Thanks for all your work on this Maya!

PR #737, opened 2025-04-22
DOCSP-32849-add-reference-to-restore-role

morales-mario-mdb:
Includes reference to the need of having the `restore` role assigned to the mongosync user for reverse sync operations in the [reference/api/start](https://www.mongodb.com/docs/cluster-to-cluster-sync/current/reference/api/start/) endpoint document.

It also includes a link to the [User Permissions](https://www.mongodb.com/docs/cluster-to-cluster-sync/current/reference/permissions/) page in the [reference/api/reverse](https://www.mongodb.com/docs/cluster-to-cluster-sync/current/reference/api/reverse/) endpoint document.

## Jira ticket

https://jira.mongodb.org/browse/DOCSP-32849

## Staging

`start` endpoint:

https://deploy-preview-737--docs-cluster-to-cluster-sync.netlify.app/reference/api/start/

`reverse` endpoint:

https://deploy-preview-737--docs-cluster-to-cluster-sync.netlify.app/reference/api/reverse/#requirements

### DIFF
--- a/source/includes/api/facts/reverse-sync.rst
+++ b/source/includes/api/facts/reverse-sync.rst
@@ -1,0 +1,3 @@
+To reverse sync, the user must have assigned the :authrole:`restore`
+role for both source and destination and the ``enableUserWriteBlocking``
+field must be set to ``sourceAndDestination``.
--- a/source/includes/fact-permissions-body.rst
+++ b/source/includes/fact-permissions-body.rst
@@ -1,5 +1,5 @@
 The user specified in the ``mongosync`` connection string must have the
-required :ref:`user permissions <c2c-permissions-and-roles>` on the source and destination
-clusters. The permissions vary depending on your environment and if you
-want to modify write-blocking settings or use reverse sync.
-
+required :ref:`user permissions <c2c-permissions-and-roles>` on the
+source and destination clusters. The permissions vary depending on your
+environment and if you want to modify write-blocking settings or use
+reverse sync.
--- a/source/includes/fact-permissions-body.rst
+++ b/source/includes/fact-permissions-body.rst
@@ -1,5 +1,5 @@
 The user specified in the ``mongosync`` connection string must have the
-required :ref:`c2c-permissions-and-roles` on the source and destination
+required :ref:`user permissions <c2c-permissions-and-roles>` on the source and destination
 clusters. The permissions vary depending on your environment and if you
 want to modify write-blocking settings or use reverse sync.
 
--- a/source/includes/fact-permissions-body.rst
+++ b/source/includes/fact-permissions-body.rst
@@ -1,5 +1,5 @@
 The user specified in the ``mongosync`` connection string must have the
-required permissions on the source and destination clusters. The
-permissions vary depending on your environment and if you want to 
-modify write-blocking settings or use reverse sync.
+required :ref:`c2c-permissions-and-roles` on the source and destination
+clusters. The permissions vary depending on your environment and if you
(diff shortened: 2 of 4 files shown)


source/includes/fact-permissions-body.rst
```diff
@@ -1,5 +1,5 @@
 The user specified in the ``mongosync`` connection string must have the
-required permissions on the source and destination clusters. The
-permissions vary depending on your environment and if you want to 
-modify write-blocking settings or use reverse sync.
+required :ref:`c2c-permissions-and-roles` on the source and destination
```
jeff-allen-mongo, line 2: [Formatting]

There's a slight issue with this approach because "user permissions" is currently rendering in title case. We should instead do something like:

```suggestion
required :ref:`user permissions <c2c-permissions-and-roles>` on the source and destination
```

This also makes sure that if the "user permissions" page title ever changes, this sentence still makes sense.

source/reference/api/start.txt
```diff
@@ -141,10 +141,14 @@ Request Body Parameters
          cluster while migration is in progress, and unblocks writes right
          before ``canWrite`` is ``true``.
 
-       To reverse sync, the ``enableUserWriteBlocking`` field must be set
-       to ``"sourceAndDestination"``. To allow the source cluster 
-       to accept writes again, for example after running migration tests, 
-       run the following command: 
+       .. include:: /includes/fact-permissions-body.rst
```
jeff-allen-mongo, line 144: I know we talked about this briefly over Zoom, but looking at this now I'm not totally sure we need this include here? Since we have this content at the top of the page, it kind of makes sense to me that we use this space to get more specific as you've done with the added info for reverse sync, but maybe we don't need to repeat the include copy? What do you think?

source/reference/api/start.txt
```diff
@@ -176,8 +180,11 @@ Request Body Parameters
      - If set to ``true``, enables the sync operation to be
        reversed.
 
-       To reverse sync, the ``enableUserWriteBlocking`` field must be set
-       to ``sourceAndDestination``.
+       .. include:: /includes/fact-permissions-body.rst
+       
+       To reverse sync, the user must have assigned the :authrole:`restore`
```
jeff-allen-mongo, line 185: Can / should we make this paragraph an include since we're using it in both the reverse and enableUserWriteBlocking descriptions?

jeff-allen-mongo (commented):
Thanks @morales-mario-mdb! Just a few questions / comments.

source/includes/fact-permissions-body.rst
morales-mario-mdb: nice, thanks!

source/reference/api/start.txt
morales-mario-mdb: Makes sense, will do 👌

source/includes/api/facts/reverse-sync.rst
```diff
@@ -0,0 +1,3 @@
+To reverse sync, the user must have assigned the :authrole:`restore`
+role for both source and destination and the ``enableUserWriteBlocking``
+field must be set to ``sourceAndDestination``.
```
jeff-allen-mongo, line 3: This is a nit, but maybe make this filename a bit more specific like `reverse-sync-requirements` or something, just in case we end up adding more `reverse` facts it's easier to distinguish.

Vodyanoy17 (approved):
LGTM

morales-mario-mdb:
Thanks @Vodyanoy17 
I'll go ahead and merge the changes

